PR #16, opened 2019-10-29
Refactor cli of josh-filter command

Nicoretti:
* Add structopt crate to simplifly cli
* Remove unused options
* Don't crash and provide a clear error message in case file does not exists
* Enable colored help
* Add additional information to help output


Nicoretti:
Do not merge -> just realised test ain't executed in your build, they 3 test fail locally -> I will fix them and push and update.

PR #135, opened 2020-11-05
fix tests ampersand redirection

LMG:


tests/proxy/setup_test_env.sh
```diff
@@ -1,5 +1,5 @@
 export TESTTMP=${PWD}
-killall josh-proxy &> /dev/null
+killall josh-proxy2 &> /dev/null
```
initcrash: this should not be committed.

LMG:
Don't forget to merge this :)

initcrash:
Changes have been squashed and applied

PR #144, opened 2020-11-17
Document duplicate file mapping

initcrash:
If a mapped path was overshadowing an already existing
path in the workspace directory, the next push from the
workspace would delete the shadowed files.
-> Fixed.


initcrash:
This breaks another feature: Moving a dir from workspace to shared location by adding to workspace file.
It's probably not fixable and some concept of workspace file verification could be a better approach.

initcrash:
As dicussed, better solution will be to print an error message explaining the situation and rejecting the push that would cause unexpected behaviour.

LMG (commented):
This is an astonishingly good changeset

LMG (approved):

PR #146, opened 2020-11-18
Multiarg

initcrash:
Multiarg support is not really new. There was already one (undocumented, mostly unused) filter called "info" that already implemented the parsing of multiple arguments for filters.
This is just making that part of the grammar and cleaning up some of the code.

### DIFF
--- a/src/filters.rs
+++ b/src/filters.rs
@@ -1253,63 +1253,59 @@ impl Filter for WorkspaceView {
 #[grammar = "view_parser.pest"]
 struct MyParser;
 
-fn make_view(cmd: &str, name: &str) -> Box<dyn Filter> {
-    if cmd == "+" || cmd == "prefix" {
-        return Box::new(PrefixView {
-            prefix: Path::new(name).to_owned(),
-        });
-    } else if cmd == "hide" {
-        return Box::new(HideView {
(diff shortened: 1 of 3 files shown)


src/filters.rs
```diff
@@ -1253,63 +1253,59 @@ impl Filter for WorkspaceView {
...
-    } else if cmd == "info" {
-        let mut s = std::collections::BTreeMap::new();
-        for p in name.split(",") {
-            let x: Vec<String> = p.split("=").map(|x| x.to_owned()).collect();
-            if x.len() == 2 {
-                s.insert(x[0].to_owned(), x[1].to_owned());
-            } else {
-                s.insert("prefix".to_owned(), x[0].to_owned());
-            }
+fn kvargs(args: &[&str]) -> std::collections::BTreeMap<String, String> {
```
LMG, line 1256: feels a bit overkill for arguments but why not

LMG (approved):
Looks good to me.
I much prefer the new make_view function.
Did I understand correctly that the actual change is the grammar, the rest is mostly cleanup?

src/filters.rs
initcrash: Yes it is :)

PR #147, opened 2020-11-18
Permissions

initcrash:


src/bin/josh-filter.rs
```diff
@@ -113,11 +124,53 @@ fn run_filter(args: Vec<String>) -> josh::JoshResult<i32> {
...
             &mut backward_maps.write().unwrap(),
         );
 
+        let mut all_dirs = vec![];
+
+        if check_permissions {
+            let result_tree = repo.find_reference(&t)?.peel_to_tree()?;
+
+            result_tree.walk(git2::TreeWalkMode::PreOrder, |_, entry| {
+                let name = entry.name().unwrap();
```
LMG, line 138: You leave a lot of unwrap()s. Is it fine to have it here? Could entry possibly be wrong? Why not unwrap_or()?

src/bin/josh-filter.rs
```diff
@@ -113,11 +124,53 @@ fn run_filter(args: Vec<String>) -> josh::JoshResult<i32> {
...
+                    );
+                    all_dirs.push(dirname);
+                }
+                git2::TreeWalkResult::Ok
+            })?;
+        }
+
+        let mut dedup = vec![];
+
+        for w in all_dirs.as_slice().windows(2) {
```
LMG, line 152: Why is all_dirs already sorted here?

src/bin/josh-filter.rs
```diff
@@ -113,11 +124,53 @@ fn run_filter(args: Vec<String>) -> josh::JoshResult<i32> {
...
+                if !b.starts_with(a) {
+                    dedup.push(a.to_owned());
+                }
+            }
+        }
+
+        if let Some(cp) = args.value_of("check-permission") {
+            let permission_regex = regex::Regex::new(cp)?;
+
+            let mut allowed = dedup.len() != 0;
```
LMG, line 163: if dedup is empty, why even do all of this?

LMG (approved):
I didn't understand everything, but apart from a few small questions it looks good to me.

src/bin/josh-filter.rs
initcrash: This is a command line tool, not a long running (server) process. So leaving some unwrap is not such a big problem here. Especially for things we don't expect to ever happen.

src/bin/josh-filter.rs
initcrash: It has been built during git2::Tree.walk(). And tree walking happens in order.

src/bin/josh-filter.rs
initcrash: Do what? If dedup is empty it only prints allowed = true and does nothing else.

PR #165, opened 2020-12-03
add parents test cases

LMG:


initcrash:
There is already a test case for this in the commit that fixes the issue.

PR #225, opened 2021-01-17
cs

initcrash:


src/filters.rs
```diff
@@ -665,11 +682,46 @@ fn last_chain(rest: Filter, filter: Filter) -> (Filter, Filter) {
...
+}
+
+/*
+ * Apply optimization steps to a filter until it converges (no rules apply anymore)
+ */
+fn optimization_iterate(filter: Filter) -> Filter {
+    let mut filter = filter;
+    loop {
+        let optimized = optimization_step(filter);
+        if filter == optimized {
```
LMG: hopefully it always converges ;D

LMG (approved):
Great changes, thanks for the documentation and comments

src/filters.rs
initcrash: Oh I had some bugs of optimization steps fighting each other resulting in infinite loop.
Hopefully got them all.
This might be a good thing to fuzz..

PR #234, opened 2021-01-21
Remove self refernce link

Nicoretti:


LMG (approved):
Moved to repo details so that a link is there in the josh github frontpage

PR #269, opened 2021-02-17
Fix filemode handling

initcrash:


LMG (approved):
I think for the workspace.josh it doesn't matter. The file will get changed by josh anyways, most of the time.

PR #303, opened 2021-03-12
change DIRS to PATHS

LMG:


src/filter/tree.rs
```diff
@@ -1,37 +1,45 @@
...
-                    &repo,
-                    &std::path::Path::new(
-                        entry.name().ok_or(super::josh_error("no name"))?,
-                    ),
-                    entry.id(),
-                    0o0100644,
-                    &result,
-                )?;
-            }
+            let file_blob = repo.blob(
```
christian-schilling, line 18: This should be: `normalize_path(std::path::Path::new(root).join(name))`

src/filter/tree.rs
```diff
@@ -1,37 +1,45 @@
...
+                    } + entry.name().ok_or(super::josh_error("no name"))?),
+                )
+                .to_str()
+                .ok_or(super::josh_error("no name"))?
+                .as_bytes(),
+            )?;
+            result = replace_child(
+                &repo,
+                &std::path::Path::new(
+                    entry.name().ok_or(super::josh_error("no name"))?,
```
christian-schilling, line 33: Also with entry.name() being used multiple times, and requireing error handling, it would make sense to make a local for it.

christian-schilling (approved):

PR #312, opened 2021-03-18
Support comments in workspace.josh files

LMG:
They start with # and are only in the first lines of the file (optionally)

### DIFF
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -5,6 +5,7 @@ mod opt;
 mod parse;
 pub mod tree;
 
+pub use parse::get_comments;
 pub use parse::parse;
 
 lazy_static! {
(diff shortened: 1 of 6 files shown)


src/filter/mod.rs
```diff
@@ -538,23 +539,26 @@ fn unapply2<'a>(
...
+            let mapped = &tree::get_blob(&transaction.repo(), &tree, &Path::new("workspace.josh"));
+            let parsed = parse(mapped)?;
+
+            let mut blob = String::new();
+            if let Ok(c) = get_comments(mapped) {
+                if c.len() > 0 {
+                    blob = format!("{}", c);
+                }
+            }
+            let blob = &format!("{}{}\n", &blob, pretty(parsed, 0));
```
christian-schilling, line 551: Will there allways be a newline at the end of the comments, or would this better be `{}\n{\n}\n`?

tests/proxy/workspace_invalid_trailing_slash.t
```diff
@@ -134,6 +134,13 @@
   remote: Can't apply "mod workspace" (4e531443c5533e6d1b2503d0fad238cfc8491807)
   remote: Invalid workspace:
   remote: ----
+  remote:  --> 1:1
+  remote:   |
+  remote: 1 | a/ = :/sub1\xe2\x90\x8a (esc)
```
christian-schilling, line 139: Where do those extra cryptic characters at the end come from, and can we get rid of them?

src/filter/mod.rs
LMG: There will, because of the grammar, I think.

LMG: Also all the newlines are kept as part of the comments. If we add one here, each push will increase the number of newlines by one.

christian-schilling (approved):

PR #328, opened 2021-04-01
Add warnings on empty mappings

LMG:


src/filter/mod.rs
```diff
@@ -446,14 +447,22 @@ fn apply2<'a>(
...
+            let res = tree::remove_pred(
                 &repo,
                 "",
                 tree.id(),
                 &|path, isblob| isblob && (pattern.matches_path_with(&path, options)),
                 git2::Oid::zero(),
                 &mut std::collections::HashMap::new(),
-            )
+            );
+            if let Some(warns) = warnings {
```
christian-schilling, line 458: I think it should be possible to do it just once for any filter, basically it's when apply() return empty_tree and then report "filter does not match anything"

christian-schilling: Maybe even doing it on a flattened compose. Then it would not pollute the rest of the code.

src/filter/mod.rs
```diff
@@ -660,6 +660,59 @@ pub fn compose(first: Filter, second: Filter) -> Filter {
...
+    transaction: &'a cache::Transaction,
+    filter: Filter,
+    tree: git2::Tree<'a>,
+) -> Vec<String> {
+    let mut warnings = Vec::new();
+
+    let tree = apply(&transaction, filter, tree);
+    if let Ok(tree) = tree {
+        if tree.is_empty() {
+            warnings.push(format!("No match for filter {}", spec(filter)));
```
christian-schilling, line 710: maybe better use `pretty()` here to have nicer output

christian-schilling (approved):

PR #331, opened 2021-04-04
cs

christian-schilling:


src/filter/tree.rs
```diff
@@ -254,6 +254,73 @@ pub fn overlay(
     return Ok(input1);
 }
 
+pub fn pathline(b: &str) -> JoshResult<String> {
+    for line in b.split("\n") {
```
LMG, line 258: don't you just want to do it for the first line? Why iterate?

christian-schilling: Because it is easier to write this way ;)

PR #347, opened 2021-04-21
Update build workflow to include all targets and tests

Nicoretti:
* Make sure all binaries and tests are built
* Fix josh-proxy unittests

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -32,13 +32,15 @@ jobs:
     - uses: actions/checkout@v2
     - name: Check format
       run: cargo fmt -- --check
-    - name: Run Rust tests
-      run: cargo test
+    - name: Run all unit tests
+      run: cargo test --workspace --all
+    - name: Build all targets (libs, binaries, examples, ..)
+      run: cargo build --workspace --all-targets
     - name: Configure environment
       run: |
         export CARGO_TARGET_DIR=`pwd`/target
         git config --global init.defaultBranch master
         git config --global user.email "christian@schilling.de"
         git config --global user.name "christian"
-    - name: Build and run tests
+    - name: Run integration tests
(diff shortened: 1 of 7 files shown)


.github/workflows/rust.yml
```diff
@@ -32,13 +32,15 @@ jobs:
...
       run: cargo fmt -- --check
-    - name: Run Rust tests
-      run: cargo test
+    - name: Run all unit tests
+      run: cargo test --workspace --all
+    - name: Build all targets (libs, binaries, examples, ..)
+      run: cargo build --workspace --all-targets
     - name: Configure environment
       run: |
         export CARGO_TARGET_DIR=`pwd`/target
```
LMG, line 41: Does this fit with building just before?

Nicoretti: I would say it may builds even more e.g. if u have examples etc.

Nicoretti: apparently previous build setup did not build josh-proxy test

LMG: I mean the CARGO_TARGET_DIR variable.
Looking in detail at the build it looks like it isn't even passed to the script...

josh-proxy/Cargo.toml
```diff
@@ -37,6 +37,6 @@ serde_json= "1.0"
 serde= "*"
 rust-crypto = "0.2"
 unindent = "*"
-juniper = "*"
+juniper = {version = "*", features =["expose-test-schema"] }
```
christian-schilling, line 40: What does that do?

LMG: exposes the feature which is now hidden behind this. The tests were not running anymore.

christian-schilling (approved):

PR #350, opened 2021-04-22
Introduce create push option

christian-schilling:


src/cache.rs
```diff
@@ -1,7 +1,7 @@
 use super::*;
 use std::collections::HashMap;
 
-const VERSION: u64 = 6;
```
LMG, line 4: on purpose?

LMG (approved):

PR #354, opened 2021-04-23
Filter UI

LMG:


josh-ui/src/patterns.rs
```diff
@@ -0,0 +1,40 @@
...
+
+pub fn path_with_note(path: &str, note: &str) -> Html {
+    html! {<>
+       {
+           if let Some(d) = std::path::Path::new(path).file_name() {
+               d.to_string_lossy().to_string()
+           } else {
+               path.to_string()
+           }
+       }{ "/" }
```
christian-schilling, line 32: This trailing slash should probably not appear everywhere, but only on directories.

christian-schilling (approved):

PR #372, opened 2021-05-03
Some Cleanups

phimuemue:
I skimmed through the code this weekend and found some cleanup opportunities.

### DIFF
--- a/src/graphql.rs
+++ b/src/graphql.rs
@@ -388,6 +388,25 @@ impl Markers {
     }
 }
 
+impl Path {
+    fn serialize_to_serde_value<E>(&self, context: &Context, str_to_value: impl FnOnce(&str)->Result<serde_json::Value, E>) -> FieldResult<Document> {
+        let transaction = context.transaction.lock()?;
+        let id = transaction
+            .repo()
+            .find_tree(self.tree)?
+            .get_path(&self.path)?
+            .id();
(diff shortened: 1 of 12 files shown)


src/graphql.rs
```diff
@@ -241,9 +241,9 @@ impl Revision {
         let commit = transaction.repo().find_commit(self.commit_id)?;
 
         let warnings = filter::compute_warnings(&transaction, self.filter, commit.tree()?)
-            .iter()
-            .map(|warn| Warning {
-                text: warn.to_string(),
+            .into_iter()
```
LMG, line 238: Out of curiosity, why is it better?

phimuemue: It avoids cloning the strings. (`iter` yields borrowed values that have to be (essentially) cloned via `to_string`, whereas `into_iter` allows for cheap moves.)

LMG: thanks!

christian-schilling (changes requested):
@phimuemue Thank you. There are some really good improvements in there.
Now it just needs to be buildable again 😅

phimuemue:
> @phimuemue Thank you. There are some really good improvements in there.
> Now it just needs to be buildable again sweat_smile

Ouch! I guess I've run `cargo check` from the wrong directory.

christian-schilling (approved):

PR #407, opened 2021-07-02
Disable UI test

christian-schilling:
Not reliable enough


LMG:
So we reenable the test when we switch to nix?

LMG (approved):

PR #411, opened 2021-07-02
Update dependencies

LMG:


LMG:
need release from tracing-opentelemetry (fix is already in)

PR #432, opened 2021-08-05
Add test showing unwanted merges in workspace

LMG:


christian-schilling:
@LMG Given that we concluded that the behavior is already correct, do you still think it makes sense to include this extra test case?
If yes I would prefer to have it converted to a `josh-filter` based test instead of proxy to make it easier and faster to run.

LMG:
Is this behaviour covered by the other tests?

LMG:
Let's drop this change. Merging should be correctly covered by the other tests.

PR #446, opened 2021-08-16
Update gettingstarted.t

christian-schilling:


LMG (approved):
Would it make sense to also run it once to update the test with more current josh output?

PR #474, opened 2021-09-12
Only connect to jaeger if env var is set

christian-schilling:
This prevents confusing errors in the log if jaeger is not used.


LMG (approved):
Needs documentation but looks good!

PR #489, opened 2021-09-21
Redirect for browser

christian-schilling:


LMG (approved):
Really nice, but you need to fix the tests :)

PR #491, opened 2021-09-25
Add shell.nix

christian-schilling:
This allows to easily setup a dev environment with all dependencies.
Would be even better with version pinning, but I don't know how to do that yet.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,19 @@
+with (import <nixpkgs> {});
+
+let
+  extra_deps = if stdenv.isDarwin then [
+    darwin.apple_sdk.frameworks.Security
+  ] else [];
+
+in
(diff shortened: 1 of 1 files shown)


LMG (commented):
This is great! works really well, helps with the cram tests!

If you want to pin each package version, we could use this tool : https://lazamar.co.uk/nix-versions/ which directly gives the command to use.

Maybe it would be enough to pin the nixpkgs version?

Nicoretti:
@christian-schilling from first glance, looks very interesting/promising I will have a closer look later today.

Nicoretti (approved):

Nicoretti:
I like it. If we can go for the version pinning as LMG mentioned perfect.

PR #495, opened 2021-10-02
shell.nix pinning, INVERT filter

LMG:


christian-schilling (approved):

bjeanes:
Is there any appetite for a PR that updates these pinned versions?

Is the team even using `shell.nix` currently? It doesn't seem functional to me. I've noticed:

- the presently pinned nixpkgs provides `rustc 1.56.1` but afaict JOSH does not compile on that version and [CI uses 1.79.0](https://github.com/bjeanes/josh/blob/a2e6df6bf9ca14976d2078d8bacd3edd5206b814/Dockerfile#L20)
- 57fbe8b5c16d74b08ac7e46aad53a3592d3c7ac8 switched from `cram` to `prysk`, but `shell.nix` is still providing `cram`
- `pkgs.nodejs-17_x` does not compile on macOS Sonoma but there are other indications in `shell.nix` that Darwin support is provided.

I'm trying to fix this locally for my own dev environment in order to propose a path forward on #1377. I could open a separate PR for `shell.nix` improvements. Alternatively, perhaps it should be removed, if its usage has been abandoned.

LMG:
You're right, I don't think shell.nix has been used since we switched to using docker https://github.com/josh-project/josh/pull/742

If you're interested in using it they it should be upgraded, but otherwise deleting it would probably be good.

PR #524, opened 2021-10-29
Enable pushing partial merge commits

christian-schilling:
While merges of branches that where made in a filtered repo
where supported for a long time. This allows merges of branches
that where created on different versions of the canonical history.
The ambiguity is resolved by picking the parent that is a descendant
of the target branch tip.

### DIFF
--- a/src/history.rs
+++ b/src/history.rs
@@ -281,7 +281,7 @@ pub fn unapply_filter(
 
         let commit_message = module_commit.summary().unwrap_or("NO COMMIT MESSAGE");
 
-        let new_trees: JoshResult<std::collections::HashSet<_>> = {
+        let new_trees: JoshResult<Vec<_>> = {
             let s = tracing::span!(
                 tracing::Level::TRACE,
                 "unapply filter",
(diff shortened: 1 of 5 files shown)


src/history.rs
```diff
@@ -311,27 +311,57 @@ pub fn unapply_filter(
...
+                } else {
+                    // This used to be our only fallback for the parent_count > 1 case.
+                    // It should never happen anymore.
+                    tracing::warn!("rejecting merge");
+                    let msg = format!(
+                        "rejecting merge with {} parents:\n{:?}",
+                        parent_count,
+                        module_commit.summary().unwrap_or_default()
+                    );
+                    return Ok(UnapplyResult::RejectMerge(msg));
```
LMG, line 363: This is the only place where "RejectMerge" is used. Why is the message fully configurable, when in other error cases they are fixed?

LMG (commented):
Looks good, is there any way we could give a warning to the user? Do we know somehow that commits are gonna get duplicated? I suppose we don't, but what about a blanket warning explaining that pushing a merge can duplicate commits?

src/history.rs
christian-schilling: Like explained in the comment, we don't expect this case to ever happen. If we still get here at some point this will give us some extra information to figure out what happened.
Actually we should probably include more in the message (at least the sha)

christian-schilling:
> Looks good, is there any way we could give a warning to the user? Do we know somehow that commits are gonna get duplicated? I suppose we don't, but what about a blanket warning explaining that pushing a merge can duplicate commits?

Whether the result of this push including a merge is what the user intended is highly context dependent. In most cases it should either be OK or caught in CI and/or code review.
So I don't thing emitting a warning here is appropriate. However, a documentation paragraph explaining the semantics of a partial merge is probably a good idea.
We could also emit a "Note" (being clear that it is not a warning but just for information) along the lines of:
"The branch you pushed contains a partial merge, read this documentation to learn what that means".

src/history.rs
LMG: I meant more the discrepancy with the other error cases, which harcode a message with an optional parameter. But I guess this is fine

LMG (approved):

PR #530, opened 2021-11-11
Pin cargo package versions in Dockerfile

vlad-ivanov-name:
Just for the sake of better reproducibility


LMG (approved):
This is good and I'll merge it, but it just made me think that I think we have a little bit of a problem regarding this stuff:

This is the dockerfile used to deploy, but we test locally using the nix-shell environment, without this docker, and on the verifier without nix-shell, or this dockerfile, depending entirely on the workflow. 

I think we need to figure out a way to keep all of this consistent

PR #543, opened 2021-11-25
Implement symref (HEAD) forwarding

christian-schilling:
Previously it was simply assumed that HEAD always points to
refs/heads/master but as it turns out that is far from true.
Now HEAD is queried per repo from the upstream and forwarded.
Unfortunately it seems that it is not avoidable to to an extra
request to the upstream server for this. (calling ls-remote)

### DIFF
--- a/josh-proxy/src/bin/josh-proxy.rs
+++ b/josh-proxy/src/bin/josh-proxy.rs
@@ -34,12 +34,15 @@ type FetchTimers = HashMap<String, std::time::Instant>;
 type Polls =
     Arc<std::sync::Mutex<std::collections::HashSet<(String, josh_proxy::auth::Handle, String)>>>;
 
+type HeadsMap = Arc<std::sync::RwLock<std::collections::HashMap<String, String>>>;
+
 #[derive(Clone)]
 struct JoshProxyService {
     port: String,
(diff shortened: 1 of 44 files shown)


josh-proxy/src/bin/josh-proxy.rs
```diff
@@ -34,12 +34,15 @@ type FetchTimers = HashMap<String, std::time::Instant>;
 type Polls =
     Arc<std::sync::Mutex<std::collections::HashSet<(String, josh_proxy::auth::Handle, String)>>>;
 
+type HeadsMap = Arc<std::sync::RwLock<std::collections::HashMap<String, String>>>;
```
LMG, line 37: Why a RwLock here, or more precisely why doesn't it also benefit the places where we use Mutex?

LMG (approved):

josh-proxy/src/bin/josh-proxy.rs
christian-schilling: You are right that a Mutex would probably work just as well given how small the critical section actually is. It is a very read heavy thing though. In the end I don't think we will be able to tell the difference with regards to performance. If any of this shared structures will ever turn out be an actual bottleneck we will need something else anyway...

PR #573, opened 2021-12-27
Fix sometimes empty headref causing errors

LMG:
I don't know how to reproduce in the tests, but I see this being a
problem in production.


christian-schilling:
Let's have a call on this one, maybe we can figure out how to reproduce.

christian-schilling (approved):

PR #575, opened 2021-12-30
Trace all HTTP codes other than 200 as errors

vlad-ivanov-name:
This allows to raise logging level but still see 500, 404 errors etc

### DIFF
--- a/josh-proxy/src/bin/josh-proxy.rs
+++ b/josh-proxy/src/bin/josh-proxy.rs
@@ -7,10 +7,11 @@ use tracing_subscriber::Layer;
 use futures::future;
 use futures::FutureExt;
 use hyper::service::{make_service_fn, service_fn};
-use hyper::{Request, Response, Server};
+use hyper::{Request, Response, Server, StatusCode};
 use std::collections::HashMap;
 use std::sync::{Arc, RwLock};
 use tokio::process::Command;
+use tracing::Span;
 use tracing_futures::Instrument;
 
 fn version_str() -> String {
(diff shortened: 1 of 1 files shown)


LMG:
Looks good but you forgot to check format!

josh-proxy/src/bin/josh-proxy.rs
```diff
@@ -678,10 +678,15 @@ async fn run_proxy() -> josh::JoshResult<i32> {
                     error_response().await
                 };
                 let _e = s.enter();
+                let http_status = r.status().as_u16()
+                let log_level = match http_status {
+                    200 => tracing::Level::TRACE,
```
christian-schilling, line 683: I would prefer something like:
```
< 400 -> TRACE // All different kinds of success
< 500 -> WARN // Something wrong with the request, clients fault
_ -> ERROR // Server side error, need to debug josh
```

christian-schilling: Actually, the first should probably be <= 401, as "unauthorized" is a very common and normal response that will happen pretty much all the time without any need for concern.

vlad-ivanov-name: @christian-schilling but it also returns 500 here when checkout URL is malformed

vlad-ivanov-name: ah I think it still makes sense. Thanks, I'll update the PR a bit later

christian-schilling: If it returns 500 when the URL is malformed that's a bug in itself, it should be something 4xx.

vlad-ivanov-name: image

christian-schilling (approved):

PR #576, opened 2021-12-30
Propagate error messages back to the client

christian-schilling:
We still need to figure out how to use the proper
status codes on a case by case basis and also make
sure the error messages will make sense to the user,
but this should already give some insight in what
is happening instead of just "there was an error"


LMG (approved):
Exactly what I wanted to do :)

PR #584, opened 2022-01-06
Improve error messages

josh-contributor:
- Print strings from JoshError
- Add a note regarding subdir filters

### DIFF
--- a/tests/proxy/clone_invalid_filter.t
+++ b/tests/proxy/clone_invalid_filter.t
@@ -53,6 +53,11 @@
 
   $ cd ${TESTTMP}
   $ git clone -q http://localhost:8002/real_repo.git:sub1.git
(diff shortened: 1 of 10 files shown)


tests/proxy/clone_invalid_filter.t
```diff
@@ -53,6 +53,11 @@
 
   $ cd ${TESTTMP}
   $ git clone -q http://localhost:8002/real_repo.git:sub1.git
-  remote: JoshError(invalid filter)
+  remote: Invalid filter command: sub1
```
christian-schilling, line 56: I would phrase is like this:
`Invalid filter: ":sub1"`
It's not really a command, and the `:` should` be included also to be more consistent with the suggested fix (which includes the `:` )

christian-schilling (approved):

PR #585, opened 2022-01-08
Improve error messages

josh-contributor:
- Handle :workspace and :prefix without arguments, and print explanations
- Handle filters not ending in .git

### DIFF
--- a/josh-proxy/src/bin/josh-proxy.rs
+++ b/josh-proxy/src/bin/josh-proxy.rs
@@ -8,6 +8,7 @@ use futures::future;
 use futures::FutureExt;
 use hyper::service::{make_service_fn, service_fn};
 use hyper::{Request, Response, Server, StatusCode};
+use indoc::formatdoc;
 use josh::JoshError;
 use std::collections::HashMap;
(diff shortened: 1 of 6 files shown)


josh-contributor:
While I can see 422 returned for filters without `.git`, the body of the response isn't shown — not sure why.

josh-proxy/src/bin/josh-proxy.rs
```diff
@@ -397,13 +398,32 @@ async fn call_service(
...
             let mut pu = parsed_url;
+
+            if pu.rest.starts_with(":") {
+                let guessed_filter = pu.rest.trim_end_matches("/info/refs");
+                return Ok(error_response(
+                    hyper::Body::from(formatdoc!(
+                        r#"
+                        Invalid filter: "{0}"
+
+                        Note: filter definitions should end with ".git":
```
christian-schilling, line 409: Note quite accurate: `.git` is not part of the filter definition, but of the url format.
So the message should be something like:
`Invalid URL: {0}`
`Note: Repository URLs should end with ".git"`

christian-schilling (approved):

PR #612, opened 2022-02-03
Add more optimiser rules for subtract 

christian-schilling:
Changing just the prefix part of a mapping in very large
workspaces resulted in very expensive filtering that can be
avoided by applying these additional rules.

### DIFF
--- a/src/filter/opt.rs
+++ b/src/filter/opt.rs
@@ -276,6 +276,22 @@ fn iterate(filter: Filter) -> Filter {
     filter
 }
 
+fn is_prefix(op: Op) -> bool {
+    match op {
+        Op::Prefix(_) => true,
+        _ => false,
+    }
+}
+
+fn prefix_of(op: Op) -> Filter {
+    let last = to_op(last_chain(to_filter(Op::Nop), to_filter(op.clone())).1);
+    to_filter(if is_prefix(last.clone()) {
+        last
+    } else {
+        Op::Nop
+    })
+}
+
 /*
  * Attempt to apply one optimization rule to a filter. If no rule applies the input
  * is returned.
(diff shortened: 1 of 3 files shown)


src/filter/opt.rs
```diff
@@ -276,6 +276,22 @@ fn iterate(filter: Filter) -> Filter {
...
 }
 
+fn is_prefix(op: Op) -> bool {
+    match op {
+        Op::Prefix(_) => true,
+        _ => false,
+    }
+}
+
+fn prefix_of(op: Op) -> Filter {
```
LMG, line 286: I'm not sure about the name here... This gets the last of a chain, if it is a prefix?

src/filter/opt.rs
```diff
@@ -347,10 +365,22 @@ fn step(filter: Filter) -> Filter {
         Op::Subtract(a, b) if a == b => Op::Empty,
         Op::Subtract(af, bf) => match (to_op(af), to_op(bf)) {
             (Op::Empty, _) => Op::Empty,
+            (_, Op::Nop) => Op::Empty,
             (a, Op::Empty) => a,
             (Op::Chain(a, b), Op::Chain(c, d)) if a == c => {
                 Op::Chain(a, to_filter(Op::Subtract(b, d)))
             }
+            (_, b) if prefix_of(b.clone()) != to_filter(Op::Nop) => {
+                Op::Subtract(af, last_chain(to_filter(Op::Nop), to_filter(b.clone())).0)
```
LMG, line 374: here you are redoing the same operation as in prefix_of. Might be optimizable?

LMG (approved):
You need to update the test and I have a few questions but otherwise looks good to me

src/filter/opt.rs
LMG: maybe last_prefix?

PR #620, opened 2022-02-06
Add rule for disjunct subtract

christian-schilling:


LMG:
Could we get some numbers on this? I'm wondering if more rules aren't making the optimization too slow

christian-schilling:
Long outdated

PR #633, opened 2022-02-14
Update rust version in dockerfile

LMG:
Apparently it was there twice, I missed it. I wonder why it's there twice


arthurbragaa (approved):
I'm approving the change, but the workflow will take much longer to run.

PR #670, opened 2022-03-26
Add include filter

LMG:
It can be inserted in workspace files, to allow for composition


christian-schilling:
A few things:
- It looks like there is quite some duplicated code from the workspace filter implementation. Could this already be factored out?
- Could the workspace filter even be expressed in terms of the include filter? (As an alias for `:[:/wsdir,:include=wsdir/workspace.josh]`
- I would prefer to not specify the `.josh` file extension in the filter, so that there is a stronger convention arround naming the files that way.

LMG:
Implemented by #1601 and #1611

PR #672, opened 2022-03-30
link traces from repo_update

LMG:
Link to the parent which is in the parent josh process

### DIFF
--- a/josh-proxy/src/bin/josh-proxy.rs
+++ b/josh-proxy/src/bin/josh-proxy.rs
@@ -2,6 +2,10 @@
 #[macro_use]
 extern crate lazy_static;
 
+use josh_proxy::RepoUpdate;
+use opentelemetry::global;
+use opentelemetry::sdk::propagation::TraceContextPropagator;
+use tracing_opentelemetry::OpenTelemetrySpanExt;
 use tracing_subscriber::Layer;
 
 use futures::future;
@@ -231,12 +235,20 @@ async fn repo_update_fn(
     let body = hyper::body::to_bytes(req.into_body()).await;
 
     let s = tracing::span!(tracing::Level::TRACE, "repo update worker");
+
     let result = tokio::task::spawn_blocking(move || {
         let _e = s.enter();
         let body = body?;
         let buffer = std::str::from_utf8(&body)?;
-        josh_proxy::process_repo_update(serde_json::from_str(buffer)?)
+        let repo_update: RepoUpdate = serde_json::from_str(buffer)?;
+        let context_propagator = repo_update.context_propagator.clone();
+        let parent_context =
+            global::get_text_map_propagator(|propagator| propagator.extract(&context_propagator));
+        s.set_parent(parent_context);
+
+        josh_proxy::process_repo_update(repo_update)
     })
+    .instrument(Span::current())
     .await?;
 
     Ok(match result {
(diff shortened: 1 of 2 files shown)


LMG:
So I suppose this does improve things a bit, although it might look a bit confusing since the repo_update_fn loses the proper parent. On the plus side, it looks like Jaeger most of the time manages to keep it together with the corresponding http_request

josh-proxy/src/bin/josh-proxy.rs
```diff
@@ -615,6 +627,15 @@ async fn call_service(
...
     }
 
+    let span = tracing::span!(tracing::Level::TRACE, "hyper_cgi");
+    let _enter = span.enter();
+    let mut context_propagator = HashMap::<String, String>::default();
+    let context = span.context();
+    global::get_text_map_propagator(|propagator| {
+        propagator.inject_context(&context, &mut context_propagator);
+    });
+    tracing::warn!("debug propagator: {:?}", context_propagator);
```
vlad-ivanov-name, line 637: leftover debug?

josh-proxy/src/bin/josh-proxy.rs
```diff
@@ -636,10 +658,8 @@ async fn call_service(
     cmd.env("JOSH_REPO_UPDATE", serde_json::to_string(&repo_update)?);
     cmd.env("PATH_INFO", parsed_url.pathinfo.clone());
 
-    let cgires = hyper_cgi::do_cgi(req, cmd)
-        .instrument(tracing::span!(tracing::Level::TRACE, "git http-backend"))
-        .await
-        .0;
+    let git_span = tracing::span!(tracing::Level::TRACE, "git http backend");
+    let cgires = hyper_cgi::do_cgi(req, cmd).instrument(git_span).await.0;
```
vlad-ivanov-name, line 662: should result of git execution go into the trace as well?

josh-proxy/src/bin/josh-proxy.rs
```diff
@@ -615,6 +627,15 @@ async fn call_service(
         }
     }
 
+    let span = tracing::span!(tracing::Level::TRACE, "hyper_cgi");
+    let _enter = span.enter();
+    let mut context_propagator = HashMap::<String, String>::default();
+    let context = span.context();
+    global::get_text_map_propagator(|propagator| {
+        propagator.inject_context(&context, &mut context_propagator);
+    });
```
vlad-ivanov-name, line 636: can this be wrapped into something with a meaningful name? like `propagate_span` or something

vlad-ivanov-name, line 636: does `global` mean that global program state is being manipulated here? would be helpful if there's a comment about that

LMG: Here we're only reading from the global configuration. This is commented on the writing side

LMG: sorry, not the writing side, but the part where I set it up. This is the part where we write in it :D

LMG: I am not sure it makes sense. From my understanding, span and _enter need to be declared in this context to link stuff properly. Maybe the three remaining lines then?

christian-schilling (approved):

PR #677, opened 2022-04-02
 Refactor CI and docker images 

josh-contributor:
* Run CI tests in docker
* Use buildx / buildkit
* Cache layers on github
* Cache rust dependencies in docker layers (when building on CI)
* Compile a specific git version from source


LMG:
superseded by #680

PR #685, opened 2022-04-07
Run filter discovery only on changed refs

christian-schilling:
Walking trees for all refs all the time caused significant resource
usage on the background job.

### DIFF
--- a/src/housekeeping.rs
+++ b/src/housekeeping.rs
@@ -1,11 +1,11 @@
 use git2;
 
 use super::*;
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::{BTreeSet, HashMap};
 use std::path::Path;
 use tracing::{info, span, Level};
 
-pub type KnownViews = BTreeMap<String, BTreeSet<String>>;
+pub type KnownViews = HashMap<String, (git2::Oid, BTreeSet<String>)>;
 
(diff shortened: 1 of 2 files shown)


src/housekeeping.rs
```diff
@@ -5,7 +5,7 @@ use std::collections::{BTreeMap, BTreeSet};
 use std::path::Path;
 use tracing::{info, span, Level};
 
-pub type KnownViews = BTreeMap<String, BTreeSet<String>>;
+pub type KnownViews = BTreeMap<String, (git2::Oid, BTreeSet<String>)>;
```
vlad-ivanov-name, line 8: I think it would be more readable if it's not just a tuple but a struct with named fields, since it's not just used for return values but fields are referenced explicitly

christian-schilling: Even better would be two separate maps and moving this to use `lazy_static` like the other in process caches.
That will be some more work though. I think this needs some rework anyway, also to store filters as `Filter` instead of `String`.
So I would say let's merge it now to see if it has the desired impact and then do those refactorings, maybe on the weekend.

vlad-ivanov-name: I see, alright. But some tests currently fail 

image

LMG (approved):

PR #697, opened 2022-04-15
Fix tests

vlad-ivanov-name:


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "josh"
-version = "0.3.0"
```
vlad-ivanov-name, line 3: this change only affects graphql, the rest was changed here https://github.com/josh-project/josh/commit/c1c55a95c98fca57529c8eeeb5ba45bad50abb29

christian-schilling (approved):

PR #708, opened 2022-04-18
Fix known filters race

christian-schilling:


src/housekeeping.rs
```diff
@@ -28,11 +33,16 @@ pub fn default_from_to(
             refs.push((refname.to_owned(), to_ref.clone()));
         }
     }
-    refs.append(&mut memorize_from_to(
-        repo,
-        &to_filtered_ref(upstream_repo, filter_spec),
-        upstream_repo,
-    ));
+
+    if filter_spec != ":/" {
```
vlad-ivanov-name, line 37: would be nice to have a comment about why ":/" filter is excluded here

src/housekeeping.rs
```diff
@@ -237,10 +245,21 @@ pub fn refresh_known_filters(
     Ok(0)
 }
 
+pub fn get_known_filters(
+) -> JoshResult<std::collections::BTreeMap<String, std::collections::BTreeSet<String>>> {
+    let mut kf = std::collections::BTreeMap::new();
+    for (k, v) in KNOWN_FILTERS.lock()?.iter() {
```
vlad-ivanov-name, line 251: can variables have more meaningful names here? or have a comment about what are k and v

vlad-ivanov-name (approved):

PR #716, opened 2022-04-24
docs: Describe most common ways of importing projects

tazjin:
Adds some initial documentation for how to import projects into a josh
repository.

There are seemingly infinitely many ways of doing this, but I have
restricted this documentation to the ones that we have found most
useful for the TVL monorepo.

cc @wpcarro @lukegb @sternenseemann

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -6,7 +6,7 @@
 # Guide
 - [Getting started](./guide/gettingstarted.md)
 - [Working with workspaces](./guide/workspaces.md)
-- [Importing projects]()
+- [Importing projects](./guide/importing.md)
 - [Integrating with CI]()
 # Reference
 - [Filter syntax](./reference/filters.md)
--- a/docs/src/guide/importing.md
+++ b/docs/src/guide/importing.md
@@ -1,0 +1,130 @@
+# Importing projects
+
+When moving to a monorepo setup, especially in existing organisations,
+it is common that the need to consolidate existing project
+repositories arises.
+
+The simplest possible case is one where the previous history of a
+project does not need to be retained. In this case, the projects files
+can simply be copied into the monoreop at the appropriate location and
+committed.
+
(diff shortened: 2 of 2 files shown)


LMG:
The way we have used in the past and that I think is simpler than the ones listed here is to add a josh repository as a remote of the repo you want to import, for example `https://josh-project.dev/josh.git:/module.git`.

You can then push to this new folder with the option `-o merge` which will preserve the history. The nice thing is that this will allow you to pull from the same url later and get the new history.

tazjin:
My understanding is that doing it that way, due to how the transforms work, will be slower on repositories that have a lot of history (as opposed to prefixing the history before merge).

We should probably present both options if we can clearly explain the benefits/drawbacks of both (we haven't used the `-o merge` option in anything other than testing, so I can't speak from practical experience).

christian-schilling:
I'm wondering what the advantage of vendoring with `git subtree` over the `josh-filter` method is even when you don't want to re-export. The only one I really see is not having to install (meaning compile from source) `josh-filter`, but this is also only really needed for an initial import with a lot of history, after that it's easier to use the proxy instead for further updates.
You never know when you want to contribute back to a vendored package.

Regarding doing the the initial import with `-o merge`: The advantage here is that you don't have to clone the entire monorepo onto your machine. I wish at some point we can make this the default approach, but right now it's more practical to do as described.

christian-schilling (changes requested):
I don't understand in which case method 2. is preferable over method 3.
Method 3. will work just fine for the one way vendoring case.
Is it *maybe* easier to use? I'm not sure.

tazjin:
Updated to drop the `git subtree` description, and add a section about pushing with `-o merge`.

I've also added a note about making commits "known" to Gerrit before merging, as this is something we've experienced both in TVL and in customer projects.

sternenseemann (approved):

christian-schilling (approved):

PR #741, opened 2022-05-18
@changes/master/christian.schilling.de@gmail.com/linear filter

christian-schilling:


docs/src/reference/filters.md
```diff
@@ -74,6 +74,14 @@ All subdirectories matching the pattern in arbitarily deep subdirectories of the
...
 All files matching the pattern in arbitarily deep subdirectories of the input
 
+## History filters
+
+These filter do not modify git trees, but instead only operate on the commit graph.
+
+### Linearize history **:linear**
+Produce a filtered history that does not contain any merge commits. This is done by
+simply dropping all parents except the first on every commit.
+
```
LMG, line 84: maybe a few words on a typical use case?

src/filter/mod.rs
```diff
@@ -61,8 +61,9 @@ enum Op {
     Nop,
     Empty,
     Fold,
-    Squash,
     Paths,
+    Squash,
+    Linear,
```
LMG, line 66: alphabetical order?

src/filter/mod.rs
```diff
@@ -274,6 +276,24 @@ fn apply_to_commit2(
...
             return Some(history::rewrite_commit(repo, commit, &[], &commit.tree()?)).transpose()
         }
+        Op::Linear => {
+            let p: Vec<_> = commit.parents().collect();
+            if p.len() == 0 {
+                return Ok(Some(commit.id()));
+            }
+            let parent = some_or!(apply_to_commit2(op, &p[0], transaction)?, {
+                return Ok(None);
+            });
```
LMG, line 286: I dont understand this case

src/filter/mod.rs
```diff
@@ -555,6 +576,7 @@ fn unapply2<'a>(
 ) -> JoshResult<git2::Tree<'a>> {
     return match op {
         Op::Nop => Ok(tree),
+        Op::Linear => Ok(tree),
```
LMG, line 579: I can see that it works, but I don't understand why. How does it find the parents after this?

LMG (commented):
It looks good to me, just a few questions to understand the implementation

src/filter/mod.rs
christian-schilling: Magic 🪄 

Just like with any other filter: When we need to find the original commits we search them by traversing the commit graph until we find `apply(sha) == filtered_sha`

src/filter/mod.rs
christian-schilling: Returning `Ok(None)` from this function means that we cannot produce filtered commit right now because one of the parents was not filtered yet. In order to not recuruse here to filter the parent we return and will get called again at a later point when the parent has been filtered.

src/filter/mod.rs
LMG: That makes sense. Thinking about it more, this didn't confuse me on other filters, but for some reason I got stuck on this one :)

LMG (approved):

PR #745, opened 2022-05-20
Remove arbitrary iteration limit

christian-schilling:
The limit of 10000 could be reached for very long histories and
produce false failure conditions.
Given that all git DAGs contain a finite number of commits
the loop should always terminate anyway.

Change-Id: for2loop


LMG (approved):
I suppose the risk is that we should never have a commit add itself to the missing array :)

PR #757, opened 2022-06-03
Execute simpler filters before more complex ones

christian-schilling:
This deterministic execution order divides the filtering of
complex filters into stages that can then be parallelised
with a later patch.

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -329,7 +329,7 @@ impl Transaction {
 
     pub fn get_missing(&self) -> Vec<(filter::Filter, git2::Oid)> {
         let mut missing = self.t2.borrow().missing.clone();
-        missing.sort();
+        missing.sort_by_key(|(f, i)| (filter::nesting(*f), *f, *i));
         missing.dedup();
         missing.retain(|(f, i)| !self.known(*f, *i));
         self.t2.borrow_mut().missing = missing.clone();
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -143,6 +143,21 @@ fn pretty2(op: &Op, indent: usize, compose: bool) -> String {
     }
 }
 
+pub fn nesting(filter: Filter) -> usize {
+    nesting2(&to_op(filter))
+}
+
+fn nesting2(op: &Op) -> usize {
+    match op {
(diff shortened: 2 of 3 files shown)


src/cache.rs
```diff
@@ -59,7 +59,7 @@ struct Transaction2 {
     path_tree: sled::Tree,
     invert_tree: sled::Tree,
     trigram_index_tree: sled::Tree,
-    missing: Vec<(filter::Filter, git2::Oid)>,
+    missing: Vec<(usize, filter::Filter, git2::Oid)>,
```
LMG, line 62: could we give this usize a name? (nesting I suppose)

src/filter/mod.rs
```diff
@@ -143,6 +143,21 @@ fn pretty2(op: &Op, indent: usize, compose: bool) -> String {
...
 
+pub fn nesting(filter: Filter) -> usize {
+    nesting2(&to_op(filter))
+}
+
+fn nesting2(op: &Op) -> usize {
+    match op {
+        Op::Compose(filters) => 1 + filters.iter().map(|f| nesting(*f)).fold(0, |a, b| a.max(b)),
+        Op::Exclude(filter) => 1 + nesting(*filter),
+        Op::Workspace(_) => usize::MAX,
```
LMG, line 154: Is this to avoid parsing workspaces?

src/filter/mod.rs
```diff
@@ -236,7 +251,14 @@ pub fn apply_to_commit(
             return Ok(id);
         }
 
-        for (f, i) in transaction.get_missing() {
+        let missing = transaction.get_missing();
+
+        let minimal_nesting = missing.get(0).map(|(x, _, _)| *x).unwrap_or(0);
```
LMG, line 256: maybe a comment saying that this assumes missing are sorted by nesting (which is true)

src/filter/mod.rs
christian-schilling: Kind of, yes. We can't parse them because we don't know the content. But it does not really matter because we don't support nested workspaces anyway, so they should always be last.

src/cache.rs
christian-schilling: Removed it ;)

LMG (approved):

PR #761, opened 2022-06-11
Don't pack refs anymore

christian-schilling:
While ref packing makes iteration a bit faster, it causes contention
when accessed concurrently, in particular when refs are deleted.


LMG:
Any idea of the impact of this on performance?

christian-schilling:
Performance we will have to observe. Could be either way, but I suspect it will be better in most cases. Never really saw an advantage with the packing.

LMG (approved):

PR #891, opened 2022-08-22
Avoid fetch for GraphQL queries with explicit sha

christian-schilling:
If the sha already exists on the proxy and no other refs are
required this gives a massive performance boost.
In particular it makes the code browser much faster.

Change-Id: graphql-no-sha-fetch

### DIFF
--- a/josh-proxy/src/bin/josh-proxy.rs
+++ b/josh-proxy/src/bin/josh-proxy.rs
@@ -461,11 +461,10 @@ async fn call_service(
         }
     };
 
-    let remote_url = [
-        serv.upstream_url.as_str(),
-        parsed_url.upstream_repo.as_str(),
-    ]
-    .join("");
+    let upstream_repo = parsed_url.upstream_repo;
+    let filter = parsed_url.filter;
+
+    let remote_url = [serv.upstream_url.as_str(), upstream_repo.as_str()].join("");
 
     if parsed_url.pathinfo.starts_with("/info/lfs") {
         return Ok(Response::builder()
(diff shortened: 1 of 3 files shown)


josh-proxy/src/bin/josh-proxy.rs
```diff
@@ -1031,9 +1016,61 @@ async fn serve_graphql(
...
         false,
     ));
-    let gql_result = josh_proxy::juniper_hyper::graphql(root_node, context.clone(), req)
-        .in_current_span()
-        .await?;
+
+    let res = {
+        let res = parsed.execute(&root_node, &context).await;
+        if !*context.allow_refs.lock().unwrap() {
+            // First atempt to serve graphql query. If we can serve it without errors
```
LMG, line 1023: attempt

josh-proxy/src/bin/josh-proxy.rs
```diff
@@ -1031,9 +1016,61 @@ async fn serve_graphql(
...
+            // That means all requested revisions where specified by SHA and we could find
```
LMG, line 1024: were

src/graphql.rs
```diff
@@ -747,6 +748,13 @@ impl RepositoryMut {
         add: Vec<MarkersInput>,
         context: &Context,
     ) -> FieldResult<bool> {
+        {
+            let mut allow_refs = context.allow_refs.lock()?;
+            if !*allow_refs {
+                *allow_refs = true;
```
LMG, line 754: It's kind of weird to me that this is done here and not up in the caller of parsed.execute (in the else). Wouldn't you risk forgetting to write it in some case down here?

LMG: nevermind, this is the error reporting actually :)

LMG (approved):

PR #894, opened 2022-08-23
Speed up history queries

christian-schilling:
History queries where very slow because `find_original(...)`
is executed on every commit.
The idea of this optimisation is reduce the amount of history
to traverse with each commit visited, taking advantage of the
topological order of the traversal.

Change-Id: history-query-optimisation

### DIFF
--- a/src/graphql.rs
+++ b/src/graphql.rs
@@ -79,6 +79,7 @@ impl Revision {
     }
 
     fn hash(&self, context: &Context) -> FieldResult<String> {
+        rs_tracing::trace_scoped!("hash");
         let transaction = context.transaction.lock()?;
         let commit = transaction.repo().find_commit(self.commit_id)?;
         let filter_commit = filter::apply_to_commit(self.filter, &commit, &transaction)?;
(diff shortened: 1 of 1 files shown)


src/graphql.rs
```diff
@@ -184,18 +186,33 @@ impl Revision {
...
+        let mut ids = {
+            rs_tracing::trace_scoped!("walk");
+            walk.skip(offset)
+                .take(limit)
+                .map(|id| id.unwrap_or(git2::Oid::zero()))
+                .collect::<Vec<git2::Oid>>()
+        };
+
+        {
+            rs_tracing::trace_scoped!("walk");
```
LMG, line 199: maybe this scope can have a different name, like find_original

src/graphql.rs
```diff
@@ -79,6 +79,7 @@ impl Revision {
     }
 
     fn hash(&self, context: &Context) -> FieldResult<String> {
+        rs_tracing::trace_scoped!("hash");
```
LMG, line 82: Did you mean to keep those here?

LMG (approved):

PR #904, opened 2022-08-25
Add split changes feature

christian-schilling:
Change-Id: split-changes

### DIFF
--- a/josh-proxy/src/lib.rs
+++ b/josh-proxy/src/lib.rs
@@ -4,7 +4,15 @@ pub mod juniper_hyper;
 #[macro_use]
 extern crate lazy_static;
 
-fn baseref_and_options(refname: &str) -> josh::JoshResult<(String, String, Vec<String>, bool)> {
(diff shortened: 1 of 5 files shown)


josh-proxy/src/lib.rs
```diff
@@ -282,6 +287,65 @@ pub fn process_repo_update(repo_update: RepoUpdate) -> josh::JoshResult<String>
...
+                        repo,
+                        &repo.find_commit(changes[i].1)?,
+                        &vec![&parent],
+                        &new_tree,
+                    )?;
+                    changes[i].1 = new_commit;
+                    new_bases.push(new_commit);
+                }
+            }
+        }
```
LMG, line 342: add break if moved.len == changes.len

christian-schilling: That has the potential for an infinite loop if there is a leftover change that cannot be rebased

LMG (approved):

PR #948, opened 2022-09-26
Implement squash filter for list of commits

christian-schilling:
This version of the squash filter will discard all but the listed commits.
The `josh-filter` command gets an option to build this list from a pattern of
refs.

### DIFF
--- a/src/filter/parse.rs
+++ b/src/filter/parse.rs
@@ -33,7 +33,8 @@ fn make_op(args: &[&str]) -> JoshResult<Op> {
             Where `path` is path to the directory where workspace.josh file is located
             "#
         ))),
-        ["SQUASH"] => Ok(Op::Squash),
+        ["SQUASH"] => Ok(Op::Squash(None)),
+        ["SQUASH", _ids @ ..] => Err(josh_error("SQUASH with ids can't be parsed")),
         ["linear"] => Ok(Op::Linear),
         ["PATHS"] => Ok(Op::Paths),
         #[cfg(feature = "search")]
--- a/tests/filter/file.t
+++ b/tests/filter/file.t
@@ -43,7 +43,7 @@
   * add file2
   * add file1
 
-  $ josh-filter -s --squash --file file.josh
+  $ josh-filter -s --single --file file.josh
   [2] :prefix=a
   [2] :prefix=b
   [3] :/sub1
--- a/tests/filter/squash.t
+++ b/tests/filter/squash.t
@@ -1,0 +1,96 @@
+  $ export RUST_BACKTRACE=1
+  $ git init -q 1> /dev/null
+
+  $ echo contents1 > file1
+  $ git add .
+  $ git commit -m "add file1" 1> /dev/null
+
+  $ git log --graph --pretty=%s
+  * add file1
+
+  $ git checkout -b branch2
+  Switched to a new branch 'branch2'
+
+  $ echo contents2 > file1
+  $ git add .
+  $ git commit -m "mod file1" 1> /dev/null
+
+  $ echo contents3 > file3
+  $ git add .
+  $ git commit -m "mod file3" 1> /dev/null
+
+  $ git checkout master
+  Switched to branch 'master'
+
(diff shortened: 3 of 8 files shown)


tests/filter/squash.t
```diff
@@ -0,0 +1,96 @@
...
+  $ git commit -m "add file2" 1> /dev/null
+
+  $ git merge -q branch2 --no-ff
+
+  $ josh-filter -s --squash "refs/tags/*" --author "New Author" --email "new@e.mail" --update refs/heads/filtered
+  Warning: reference refs/heads/filtered wasn't updated
+  [1] :SQUASH=e69de29bb2d1d6434b8b29ae775ad8c2e48c5391
+
+  $ git log --graph --decorate --pretty=oneline refs/heads/filtered
+  fatal: ambiguous argument 'refs/heads/filtered': unknown revision or path not in the working tree.
```
LMG, line 36: Is this expected?

tests/filter/file.t
```diff
@@ -43,7 +43,7 @@
   * add file2
   * add file1
 
-  $ josh-filter -s --squash --file file.josh
+  $ josh-filter -s --single --file file.josh
```
LMG, line 46: now that I think about it, it's weird that the result of this is the "initial" commit and not one of the three above...

tests/filter/squash.t
LMG: ah yes cause you don't have tags yet, stupid me

tests/filter/squash.t
```diff
@@ -0,0 +1,96 @@
...
+
+  $ echo contents3 > file2
+  $ git add .
+  $ git commit -m "add file2" 1> /dev/null
+
+  $ git merge -q branch2 --no-ff
+
+  $ josh-filter -s --squash "refs/tags/*" --author "New Author" --email "new@e.mail" --update refs/heads/filtered
+  Warning: reference refs/heads/filtered wasn't updated
+  [1] :SQUASH=e69de29bb2d1d6434b8b29ae775ad8c2e48c5391
```
LMG, line 33: weird that it squashes something anyways

tests/filter/squash.t
```diff
@@ -0,0 +1,96 @@
...
+  * 0b4cf6c9efbbda1eada39fa9c1d21d2525b027bb (tag: tag_b) add file1
+
+  $ josh-filter -s --squash "refs/tags/*" --author "New Author" --email "new@e.mail" --update refs/heads/filtered
+  [1] :SQUASH=e69de29bb2d1d6434b8b29ae775ad8c2e48c5391
+  [2] :SQUASH=e8e83b9c5d2f779f0cea83a6cad68b710a399c96
+  [3] :SQUASH=3953063f3dc58661e9db16f9014aab1e8ec50bf8
+
+  $ git log --graph --decorate --pretty=oneline refs/heads/filtered
+  * 5b1a753860ca124024f6dfb4fd018fe7df8beae4 (tag: filtered/tag_a, filtered) refs/tags/tag_a
+  |\
```
LMG, line 66: Is this a merge commit? If so what with?

src/filter/parse.rs
```diff
@@ -33,7 +33,8 @@ fn make_op(args: &[&str]) -> JoshResult<Op> {
             Where `path` is path to the directory where workspace.josh file is located
             "#
         ))),
-        ["SQUASH"] => Ok(Op::Squash),
+        ["SQUASH"] => Ok(Op::Squash(None)),
+        ["SQUASH", _ids @ ..] => Err(josh_error("SQUASH with ids can't be parsed")),
```
LMG, line 37: what does this mean?

tests/filter/squash.t
christian-schilling: It does not. This just means that there is an entry in the table.

tests/filter/squash.t
christian-schilling: yes, because the original is a merge commit. The overall topology (merges) is maintained with this filter. That is the point of it ;)

Issue #957, opened 2022-10-01
Imperfect round-trip of Miri repo due to gpgsig


RalfJung:
When applying josh to https://github.com/rust-lang/miri, doing a round-trip that *should* be the identity nevertheless produces a result with a different hash:
```
git clone https://github.com/rust-lang/miri
cd miri
# do this in 2 steps because otherwise josh is too clever and doesn't do any work
josh-filter :prefix=extra refs/heads/master --update refs/heads/filtered
josh-filter :/extra refs/heads/filtered --update refs/heads/double-filtered
git diff refs/heads/master refs/heads/double-filtered
# no output, as expected
git rev-parse refs/heads/master refs/heads/double-filtered
# two different SHAs
```
I went through our history to see where the difference starts, and it is [this commit](https://github.com/rust-lang/miri/commit/6abfa56b20009546be67837cc7cd92c9c13f5951). `show --pretty=raw` also shows why:
```
commit 6abfa56b20009546be67837cc7cd92c9c13f5951
tree b450b00d80b565f5ef7f93660af60468b6add237
parent 8cac01ae66a8eebca4711a5771ad161cb8f498ba
author Pascal Hertleif <killercup@gmail.com> 1460584860 +0200
committer Pascal Hertleif <killercup@gmail.com> 1460584860 +0200
gpgsig -----BEGIN PGP SIGNATURE-----
 Comment: GPGTools - http://gpgtools.org
 
 iEYEABECAAYFAlcOwccACgkQ7bsajSBHoHQp+gCgqi423y1LEedj4qkVGvdzGx28
 OXQAnijThOzObWreOk0wRMZ8RvTE7/EK
 =q0is
 -----END PGP SIGNATURE-----

    Update to Rust Nightly 2016-04-11

diff --git a/src/interpreter.rs b/src/interpreter.rs
index 850ddf8aa18..a1db521760f 100644
--- a/src/interpreter.rs
+++ b/src/interpreter.rs
@@ -1,7 +1,7 @@
 use arena::TypedArena;
 use rustc::infer;
 use rustc::middle::const_val;
-use rustc::middle::def_id::DefId;
+use rustc::hir::def_id::DefId;
 use rustc::mir::mir_map::MirMap;
 use rustc::mir::repr as mir;
 use rustc::traits::{self, ProjectionMode};
```
The `gpgsig` part is lost after josh did its thing.

[...]

PR #958, opened 2022-10-01
preserve commit signatures

RalfJung:
Note that I don't really know what I am doing here. ;) This seems like a massive breaking change to me -- if someone's git history contains these signatures, now the commits exposed via josh will be all different. Is that a problem? What is the usual way to deal with such a problem? Should there be a flag? Were signatures actually dropped deliberately? So many questions.^^

The first commit is just me learning to write these tests, but I couldn't find an existing test specifically for the prefix filter so I figured I might submit it as well. If you prefer I can remove it.

Fixes https://github.com/josh-project/josh/issues/957


RalfJung:
For instance, the docs for importing say
```
4. Any external users can now use the `:/tools/project-b` josh filter
   to retrieve changes made in the new project location - without the
   git hashes of their existing commits changing (that is to say,
   without conflicting).
```
but currently this is only true if the original project did not contain any signed commits.

src/history.rs
```diff
@@ -196,6 +196,18 @@ pub fn rewrite_commit(
         parents,
     )?;
 
+    if let Ok((sig, _)) = repo.extract_signature(&base.id(), None) {
+        // Re-create the object with the original signature (which of course does not match any
+        // more, but this is needed to guarantee perfect round-trips).
```
RalfJung, line 201: FWIW there is a comment a few lines up that needs updating now -- or we could entirely remove that equality check.

src/history.rs
```diff
@@ -196,6 +196,18 @@ pub fn rewrite_commit(
         parents,
     )?;
 
+    if let Ok((sig, _)) = repo.extract_signature(&base.id(), None) {
```
RalfJung, line 199: `None` here and in the `commit_signed` below refers to the name of the signature field. Looks like that can be customized. Roundtrips will still not work if there are commits that use a non-standard name for that field. I don't know how to get git to tell us the name of that field... really ideally we'd have an API to list all header fields, and preserve them all while only adjusting the `tree`, but I am not sure if that is possible.

christian-schilling: That's quite interesting: We could maybe at some point even use that for out own metadata. Good to know.

About getting all extra header fields: Worst case we have to parse the commit buffer ourselves. But I think this is good enough for now.

One little thing: The constant `CACHE_VERSION` in `src/cache.rs` should be incremented whenever we make a change that can affect the output sha's of any filter. This is to avoid strange effects that could happen when somebody uses a cache that was computed using the old version and things could get mixed up.

src/history.rs
christian-schilling: I think that equality check can be removed. It's value as an optimization is super low as it is an quite rare case.

christian-schilling (approved):

Issue #956, opened 2022-10-01
Missing leading spaces in cram test?


RalfJung:
I am trying to learn cram testing by example to add a test to josh, and I am confused by [this test](https://github.com/josh-project/josh/blob/5b281ed98c6a8e29fdb5d42a1151d4f491b23aa1/tests/filter/submodule.t#L35). If I understand the cram syntax properly, things not starting with two spaces are comments. But that file has a bunch of commands that look like they should be tested, but don't start with two spaces, and hence these are comments? Is that a bug in the test or am I misunderstanding something?

christian-schilling:
You are understanding everything right.
Looking at the history of this I remember: Those lines where never really part of the test. It was sort of a "draft" of something to implement but that never happened.
So not the best test case to learn from ;)

RalfJung:
Ah, that is bad luck. Thanks for confirming. :)

PR #959, opened 2022-10-01
remove leftover draft commands from test

RalfJung:
Fixes https://github.com/josh-project/josh/issues/956


christian-schilling:
@vlad-ivanov-name I think we can then just delete the whole test case or?

christian-schilling (approved):

Issue #952, opened 2022-09-29
Josh generates lots of redundant merge commits


RalfJung:
To reproduce
```
docker run -p 8000:8000 -e JOSH_REMOTE=https://github.com -v josh-vol:/data/git joshproject/josh-proxy:latest
```
Then in another terminal
```
git clone https://github.com/rust-lang/miri
cd miri
git remote add josh http://localhost:8000/rust-lang/rust.git:/src/tools/miri.git
git fetch josh
git merge josh/master
```
The resulting history duplicates *all* merge commits from the rustc repo, which is not pretty and makes me wonder why -- most of them have nothing to do with this subrepo:

![image](https://user-images.githubusercontent.com/330628/193000916-60116795-f5a5-47b5-9a01-2ba5790c557f.png)

Cc @oli-obk

christian-schilling:
So I tried it myself:
<img width="180" alt="Screenshot 2022-09-29 at 15 52 07" src="https://user-images.githubusercontent.com/15813/193050612-a7bca1a8-0206-4c20-98aa-95199354fcf3.png">

This is the entire history of `/src/tools/miri` so I don't think that those are *all* the merge commits from the main rust repo.

The reason that those commits are kept even though they look redundant:
When adding an external repo as a subdirectory into a monorepo (what you want to do) Josh guarantees that splitting that subdirectory back out will yield the exact same sha1's like the original repo. (Most, if not all, other filtering tools do not make that guarantee)
It took quite a lot of tweaks to the commit selection rules to handle all the corner cases of wired original histories (people do strange things...) and one of those rules that proved important is:
"Never change the number of parents a commit has"
In other words: If an original commit is a merge commit, the filtered one has to be a merge commit as well.
A second (more obvious) rule is: "Keep all commits that differ from at least *one* of their parents"
Git usually only shows the diff to the first parent when looking at history, so some of those commits may look redundant, but they have a non empty diff with their second parent.

RalfJung:
Thanks for taking a look!

>  those are all the merge commits from the main rust repo.

All the ones since we did the first josh sync.... or so I thought, but yeah it seems to be only the ones on the path from some Miri-changing commit to master. That's still quite a few, and it will be one merge commit for each PR that lands in rustc (the vast majority of which does not change Miri).

> The reason that those commits are kept even though they look redundant:

[...]

Issue #950, opened 2022-09-28
Cannot push to new branch on remote


oli-obk:
I am trying to push changes from my partial fork to the main fork and then open a PR against the main fork. I am getting odd git errors I've never seen before:

```
ubuntu@ip-172-31-35-201 ~/c/miri (josh)> git push -u josh-oli josh
Enumerating objects: 40345, done.
Counting objects: 100% (40345/40345), done.
Delta compression using up to 96 threads
Compressing objects: 100% (13648/13648), done.
Writing objects: 100% (40345/40345), 9.54 MiB | 6.07 MiB/s, done.
Total 40345 (delta 28201), reused 37473 (delta 25565)
remote: Resolving deltas: 100% (28201/28201), done.
remote: josh-proxy
remote: response from upstream:
remote: Reference "refs/heads/josh" does not exist on remote.
remote: If you want to create it, pass "-o base=<basebranch>" or "-o base=path/to/ref"
remote: to specify a base branch/reference.
remote: 
remote: 
remote: 
remote: error: hook declined to update refs/heads/josh
To http://localhost:8000/oli-obk/rust.git
 ! [remote rejected]   josh -> josh (hook declined)
```

Is there a better way to sync changes back?

oli-obk:
I have tried creating a branch and pushing to that, but no matter which commit I place that branch on, I always get

```
hint: Updates were rejected because the remote contains work that you do
hint: not have locally. This is usually caused by another repository pushing
hint: to the same ref. You may want to first integrate the remote changes
hint: (e.g., 'git pull ...') before pushing again.
```

no matter which commit I place the branch on the remote at. Is there a specific commit I need to point it at?

LMG:
Hi @oli-obk , 

what the first message was trying to tell you is that josh doesn't know *where* to put this new branch you're pushing. So you should tell it, by providing a "base", another branch from which yours was derived. You can do that with 
```
git push origin your-branch -o base=master
```
for example, if you branched off of master.

Now that you've created your remote branch, you might want to delete it and try this way instead.

christian-schilling:
Hi @oli-obk,
When creating a new branch via a partial repo, you need to pass the base (target) branch when doing the push. It says so in the error message.
So if your work is based on `refs/heads/master` and you want to create a PR to merge into `refs/heads/master` you need to push like this:
`git push -o base=refs/heads/master origin local_branch_name:refs/heads/remote_branch_name`

oli-obk:
[...]

PR #961, opened 2022-10-02
Add support for subtree-conditional prefix filter

RalfJung:
This helps with the situation where a subfolder was merged into a monorepo via `git subtree`, but now one wants to use josh instead but re-use the existing history in the monorepo. `git subtree` works different than josh in the sense that the monorepo contains the original unaltered commits from the subrepo, and then links them up with the remaining monorepo history using merge commits. To get josh to extract the desired history from that, we need a way to tell josh to treat the subrepo commits as being in the subrepo -- that is basically the `prefix` filter, moving the commit contents into the appropriate folder, but only for those commits that come from the subrepo.

This PR achieves that with a new `:subtree_prefix=commit,path` filter. Here `commit` should be the last subrepo commit that was merged into the monorepo using `git subtree`; that commit and all its ancestors will then be treated with the `:prefix=path` filter while all other commits will be passed through unaltered.

To implement that I had to propagate commit information down to the filter implementation; I hope the way I did that makes sense. I am slightly worried that there might be caching somewhere that says "oh I already processed this tree with that filter, let me re-use the old result" -- which might be invalid now if different commits share the same tree, because the new filter is commit-dependent. I also don't quite know what to do in `unapply_filter`, but making this filter a NOP there made the `--reverse` part of the test pass.

I have verified that I can extract the Miri history from the rustc repo with this, using
```
git fetch http://localhost:8000/rust-lang/rust.git:subtree_prefix=75dd959a3a40eb5b4574f8d2e23aa6efbeb33573,src/tools/miri:/src/tools/miri.git master
```
and (if I also include https://github.com/josh-project/josh/pull/958) this will extract a history that perfectly reproduces Miri until commit 75dd959a3a40eb5b4574f8d2e23aa6efbeb33573. I still have to test pushing from the Miri repo to the josh remote to re-integrate changes in rustc.

Fixes https://github.com/josh-project/josh/issues/950
Fixes https://github.com/josh-project/josh/issues/952


src/filter/opt.rs
```diff
@@ -418,6 +418,14 @@ pub fn invert(filter: Filter) -> JoshResult<Filter> {
         Op::Subdir(path) => Some(Op::Prefix(path)),
         Op::File(path) => Some(Op::File(path)),
         Op::Prefix(path) => Some(Op::Subdir(path)),
+        Op::SubtreePrefix {
+            subtree_tip: _,
+            path: _,
+        } => {
+            // We assume that new commits being added don't need the prefix fixup any more.
+            // FIXME(RalfJung): does that make sense?
```
RalfJung, line 426: The alternative would be to do the ancestor test as well here. Not sure if that makes sense though, it would only fire if the commits line up perfectly -- I don't quite know what even the commit ID that gets passed down here. (Or rather, currently it doesn't get passed down yet, the callers would have to propagate that.)

christian-schilling:
Wow that was quick. Thanks for taking a stab at this!
I did some very initial (not yet working) prototyping for this as well and I think we don't need the `is_ancester_of` function:
When implementing the filter in `apply_to_commit` instead of `apply(tree)` all that really matters is the one commit that matches the given sha. For this one the prefix filter has to be used (wich will automatically transform it's parents) and all for all other commits it's `Nop`.

I have an open pr #948 that also introduces a first filter that has no implementation of `apply(tree)` but only `apply_to_commit`. That being said I'm also considering if your idea of passing the commit to `apply(tree)` might be a good thing after all...not sure yet.

Regarding the exposed API of this feature: I also have (very different) use case for this "filter that applies only to parts of the history" feature, but not with the `:prefix` filter. So I think the better approach here is to make a more generic filter like `:start=sha[:filter]` (in your case that would mean `:start=75dd959a3a40eb5b4574f8d2e23aa6efbeb33573[:prefix=src/tools/miri]`).
That way we gain more flexibility in using that new feature.

RalfJung:
> Regarding the exposed API of this feature: I also have (very different) use case for this "filter that applies only to parts of the history" feature, but not with the :prefix filter. So I think the better approach here is to make a more generic filter like :start=sha[:filter] (in your case that would mean :start=75dd959a3a40eb5b4574f8d2e23aa6efbeb33573[:prefix=src/tools/miri]).
That way we gain more flexibility in using that new feature.

Yeah that definitely makes a lot of sense. I didn't want to get into higher-order filters or changing the filter grammar though. ;)

I'm totally fine with basically all of this code being thrown away. What I care about is that josh can express these kind of history transformations, not the implementation details.

> I did some very initial (not yet working) prototyping for this as well and I think we don't need the is_ancester_of function:
When implementing the filter in apply_to_commit instead of apply(tree) all that really matters is the one commit that matches the given sha. For this one the prefix filter has to be used (wich will automatically transform it's parents) and all for all other commits it's Nop.

I don't understand how this can work, but I also don't understand josh very well. We want that commit and all of its parents to be extracted, after all, so how can it be enough to transform one commit? Or is the point that `apply_to_commit` is anyway recursively descending the commit chain and that's how the ancestor information is implicitly propagated? That sounds much more elegant indeed.

I can try your prototype on our history and tell you if the results are anything like what we need. ;) Let me know once it is pushed somewhere.

RalfJung:
Also something odd seems to be happening when pushing to such a filter. It worked fine for pushing a tiny change, but now I am pushing a larger and more messy history and it is extremely slow. The only thing I see in the log is every so often a new 
```
INFO http_request:call_service:hyper_cgi:http_request:call_service:hyper_cgi:http_request:call_service:hyper_cgi:http_request:call_service:repo_update_fn:repo update worker:unapply_filter
```
Is that a message that appears for each pushed commit? That would put it to around 10-30s for each commit. Which is surprising giving that fetching and transforming the entire large rustc history on a pull only takes a few minutes.

RalfJung:
Ah, that was a push affected by https://github.com/josh-project/josh/issues/950 -- pushing things without the bad josh sync that happened into the Miri repo works fine.

christian-schilling:
Here is my implementation as "higher order filter" it passes the test case from this PR.

PR #965, opened 2022-10-03
Implement the `:rev(...)` filter

christian-schilling:
This allows selectively applying filtering rules to different parts of the history by providing a sha -> filter mapping as an argument.


RalfJung:
I am trying this on Miri, and something doesn't seem right...
```
git fetch http://localhost:8000/rust-lang/rust.git:start=75dd959a3a40eb5b4574f8d2e23aa6efbeb33573[:prefix=src/tools/miri]:/src/tools/miri.git master
```
I started this a few hours ago and completely forgot that it is still running. Clearly it is not going anywhere.

The corresponding command with https://github.com/josh-project/josh/pull/961 is
```
git fetch http://localhost:8000/rust-lang/rust.git:subtree_prefix=75dd959a3a40eb5b4574f8d2e23aa6efbeb33573,src/tools/miri:/src/tools/miri.git master
```
and it completes in around 10min on an empty cache.

christian-schilling:
> I started this a few hours ago and completely forgot about it, and it is still running. Clearly it is not going anywhere.

Found the bug, fixed in newest version of this PR.

The reason was in contrast to the test case, in the real world you used a chained filter (`:/a/b` is internally converted to `:/a:/b`) and those chained filters are not cached directly. The implementation relied on finding the `startfilter` in the cache, which never happened for such a filter.

Now this should work (adapted test case, and tried it on miri as well).

The only question remaining for me is if `start` is really the final name for this filter ;)

RalfJung:
Yeah I was wondering about that name. `:if_ancestor_of=sha[filter]` or something like that maybe? I feel like it should invoke a conditional, to make clear that it doesn't just extract that subhistory.

RalfJung:
The new branch looks great. :)

I tried pulling from Rust, and the history contained commit 338c50e396f31c3970ab4ae7bc5d79807c6c033b which is what I extracted with my branch.

I trued pushing to Rust, and got commit 6f2b52ff103f9095fb2817637bc7d388bd4c7f6b, the exact same as with by branch. This includes the strange choice of base commit, 8e3b9bca65d7d79a3b0e9c33fed8d8c93dd66041.

RalfJung:
`at_commit` seems fine, though I would think that this applies only to that commit, not to the entire subhistory starting at that commit.

christian-schilling:
>  though I would think that this applies only to that commit, not to the entire subhistory starting at that commit.

It does only apply to that one specific commit, not the subhistory before it. That was actually a bug in the other implementation. Consider a history like this:

```
* A
|  * B
|  * C
| /
* D
* E
```
And the filter `:at_commit=C[:filter]`
Now how will the filtered versions `A'`and `B'` of `A` and `B` look like?
The filtered history of `A` will be just `A - D - E` while `B` will be `B - C' - D' - E'`.
The previous implementation would, if `B` was filtered first insert filtered commits for `D` end `E` into the cache and then deliver wrong results for `A`.

So the presence of C further up the history does not in any case change how `D` and `E` are filtered. It only affects `C`. However because `C` is "substituted" with `C'` and the parent of that is `D'` it looks like the previous history was filtered. This is kind of "switching lanes" and not like actually changing the filter on the same lane.

RalfJung:
> It does only apply to that one specific commit, not the subhistory before it. 

But what about `:at_commit=B[:filter]`? Both B and C are being filtered now. So it's definitely not just at that commit.

It is true that I did not consider the situation where the subtree "merges back" into the main history. For the use-cases I can imagine for this filter, that will never happen. But it is also true that more than 1 commit is being filtered -- all the commits of the `git subtree`-imported Miri history in rustc are being filtered.

christian-schilling:
Yes, they are being filtered kind of as a "dependency" of the `at_commit` filter. Even on a perfectly linear history, the filtered versions of the commits before the "chosen" will not look filtered if you ask for them directly but you will get a history that includes the filtered versions if you ask for a commit further up.

The reason I have not merged this yet, is because I keep thinking about what to do in case you want multiple conditional filter operations in combination. Chaining will not really work because you don't know the sha in advance.
I'm considering a filter that accepts multiple (sha, filter) pairs, but for that the grammar will need to be extended even more. 🤔 
Might still be worth it though.

RalfJung:
> Yes, they are being filtered kind of as a "dependency" of the at_commit filter. Even on a perfectly linear history, the filtered versions of the commits before the "chosen" will not look filtered if you ask for them directly but you will get a history that includes the filtered versions if you ask for a commit further up.

Okay that is pretty confusing. From a user perspective it sure looks like the filter just applies to all these commits.

(Also I think in my branch they would have looked filtered, if the "chosen" commit already exists so that the ancestor check applies.)

> The reason I have not merged this yet, is because I keep thinking about what to do in case you want multiple conditional filter operations in combination. Chaining will not really work because you don't know the sha in advance.

Yeah I wondered about that too -- it might happen in rustc if other tools also switch to josh. My implementation propagated the same commit id to every filter in the chain so I think it would have worked with chaining, I have no idea what yours will do when `at_commit` is not the first filter in the chain.

christian-schilling:
Propagating the original commit down the chain would require also adding that to the key used for caching. Making the cache a lot bigger potentially and also making hits less likely.
It would break some of the basic ideas that make josh as efficient as it es today.

So I think in the end this special filter will have to be that: special. We already have that in another case: The workspace filter can only be in the top level chain and not anywhere deeper.

This will be similar.

However I think I want to solve the "multiple chosen commits" situation before settling on an API for this.

I'm confident I though that shas will remain compatible so you can already use this as is if you don't mind running a non released version for some time.

RalfJung:
Yeah I can see how propagating the commit could break a data model.

> I'm confident I though that shas will remain compatible so you can already use this as is if you don't mind running a non released version for some time.

Sounds good, thanks!

RalfJung:
I am now encountering a strange error trying to fetch commits with this version of josh
```
$ git fetch http://localhost:8000/rust-lang/rust.git:at_commit=75dd959a3a40eb5b4574f8d2e23aa6efbeb33573[:prefix=src/tools/miri]:/src/tools/miri.git master
remote: parse_item: no match
fatal: unable to access 'http://localhost:8000/rust-lang/rust.git:at_commit=75dd959a3a40eb5b4574f8d2e23aa6efbeb33573[:prefix=src/tools/miri]:/src/tools/miri.git/': The requested URL returned error: 500
```
I started josh via
```
cargo run --release -p josh-proxy -- --local=$(pwd)/local --remote=https://github.com --no-background
```
The josh log is
```
2022-10-09T10:08:59.921219Z DEBUG http_request:call_service:hyper_cgi:http_request:call_service:hyper_cgi:http_request:call_service: josh_proxy: static_path "/rust-lang/rust.git:at_commit=75dd959a3a40eb5b4574f8d2e23aa6efbeb33573[:prefix=src/tools/miri]:/src/tools/miri.git/info/refs" path="/rust-lang/rust.git:/src/tools/miri.git/git-upload-pack" serv=JoshProxyService { repo_path: "/home/r/src/rust/josh/local", upstream_url: "https://github.com" } req_auth=(Handle { value: "" }, Request { method: POST, uri: /rust-lang/rust.git:/src/tools/miri.git/git-upload-pack, version: HTTP/1.1, headers: {"host": "localhost:8000", "user-agent": "git/2.35.1", "accept-encoding": "deflate, gzip, br, zstd", "content-type": "application/x-git-upload-pack-request", "accept": "application/x-git-upload-pack-result", "content-length": "960"}, body: Body(Streaming) }) path="/rust-lang/rust.git:/src/tools/miri.git/git-upload-pack" serv=JoshProxyService { repo_path: "/home/r/src/rust/josh/local", upstream_url: "https://github.com" } req_auth=(Handle { value: "" }, Request { method: POST, uri: /rust-lang/rust.git:/src/tools/miri.git/git-upload-pack, version: HTTP/1.1, headers: {"host": "localhost:8000", "user-agent": "git/2.35.1", "accept-encoding": "deflate, gzip, br, zstd", "content-type": "application/x-git-upload-pack-request", "accept": "application/x-git-upload-pack-result", "content-encoding": "gzip", "content-length": "822"}, body: Body(Streaming) }) path="/rust-lang/rust.git:at_commit=75dd959a3a40eb5b4574f8d2e23aa6efbeb33573[:prefix=src/tools/miri]:/src/tools/miri.git/info/refs" serv=JoshProxyService { repo_path: "/home/r/src/rust/josh/local", upstream_url: "https://github.com" } req_auth=(Handle { value: "" }, Request { method: GET, uri: /rust-lang/rust.git:at_commit=75dd959a3a40eb5b4574f8d2e23aa6efbeb33573[:prefix=src/tools/miri]:/src/tools/miri.git/info/refs?service=git-upload-pack, version: HTTP/1.1, headers: {"host": "localhost:8000", "user-agent": "git/2.35.1", "accept": "*/*", "accept-encoding": "deflate, gzip, br, zstd", "accept-language": "en-US, *;q=0.9", "pragma": "no-cache", "git-protocol": "version=2"}, body: Body(Empty) })
2022-10-09T10:08:59.921307Z ERROR http_request:call_service:hyper_cgi:http_request:call_service:hyper_cgi:http_request: josh_proxy: http_status=500 path="/rust-lang/rust.git:/src/tools/miri.git/git-upload-pack" serv=JoshProxyService { repo_path: "/home/r/src/rust/josh/local", upstream_url: "https://github.com" } req_auth=(Handle { value: "" }, Request { method: POST, uri: /rust-lang/rust.git:/src/tools/miri.git/git-upload-pack, version: HTTP/1.1, headers: {"host": "localhost:8000", "user-agent": "git/2.35.1", "accept-encoding": "deflate, gzip, br, zstd", "content-type": "application/x-git-upload-pack-request", "accept": "application/x-git-upload-pack-result", "content-length": "960"}, body: Body(Streaming) }) path="/rust-lang/rust.git:/src/tools/miri.git/git-upload-pack" serv=JoshProxyService { repo_path: "/home/r/src/rust/josh/local", upstream_url: "https://github.com" } req_auth=(Handle { value: "" }, Request { method: POST, uri: /rust-lang/rust.git:/src/tools/miri.git/git-upload-pack, version: HTTP/1.1, headers: {"host": "localhost:8000", "user-agent": "git/2.35.1", "accept-encoding": "deflate, gzip, br, zstd", "content-type": "application/x-git-upload-pack-request", "accept": "application/x-git-upload-pack-result", "content-encoding": "gzip", "content-length": "822"}, body: Body(Streaming) }) path="/rust-lang/rust.git:at_commit=75dd959a3a40eb5b4574f8d2e23aa6efbeb33573[:prefix=src/tools/miri]:/src/tools/miri.git/info/refs"
```
This is so verbose that it is pretty useless to me.^^

Fetching without the `at_commit` filter works fine (`git fetch http://localhost:8000/rust-lang/rust.git:/src/tools/miri.git master`). I tried deleting the `local` folder, which did not help.

RalfJung:
Oh, I think I was on an outdated version of this branch. I guess the `remote: parse_item: no match` meant that I had an error in my filter expression. I wasn't sure whether that error came from git or josh.

christian-schilling:
So it works now?

RalfJung:
Yeah, sorry for the noise. Only the parse error message could be improved. :)

docs/src/reference/filters.md
```diff
@@ -91,6 +91,13 @@ These filter do not modify git trees, but instead only operate on the commit gra
 Produce a filtered history that does not contain any merge commits. This is done by
 simply dropping all parents except the first on every commit.
 
+### Filter specific part of the history **:at_commit=<sha>[:filter]**
+Produce a history where the commit specified by `<sha>` is replaced by the result of applying
+`:filter` to it.
+This means also all parents of this specific commit appear filtered with `:filter` and all
+descendent commits will be left unchanged. However all commits hashed will still be different
```
sternenseemann, line 98:
```suggestion
descendent commits will be left unchanged. However all commits hashes will still be different
```

RalfJung:
Strangely, I am now getting an error when trying to push to this filter. The same thing worked fine many times before, but now when I try to push the latest bit of Miri history it fails.

From a clone of https://github.com/rust-lang/miri, I am doing
```
git push http://localhost:8000/RalfJung/rust.git:at_commit=75dd959a3a40eb5b4574f8d2e23aa6efbeb33573[:prefix=src/tools/miri]:/src/tools/miri.git -o base=master HEAD:miri
```
In `RalfJung/rust`, `miri` does not exist and `master` is [b1ab3b7](https://github.com/RalfJung/rust/commit/b1ab3b738ac718da74cd4aa0bb7f362d0adbdf84), the same commit that the last josh pull was made from.

Pushing then shows
```
remote: josh-proxy
remote: response from upstream:
remote: rejecting merge with 2 parents:
remote: "Auto merge of #2583 - RalfJung:rustup, r=oli-obk"
```
That can only be referring to 85f449aafb9c9a522e03b9d9a05e0d7b47674586, which was already present in the last mergers and did not cause problems before. I don't understand what is different now.

RalfJung:
FWIW I am getting the same error with my own subtree_filter branch. I am sure I tested something like this before we decided to land the josh changes, but clearly what I tested was not representative for an actual `push` with real changes in both repos. (I have first synced the monorepo changes into the subrepo, and now I am failing to push the resulting sync back to the monorepo.)

RalfJung:
I managed to create a [reasonable-looking sync](https://github.com/rust-lang/rust/pull/103392) with the following hack
```diff
diff --git a/src/history.rs b/src/history.rs
index 299d05b..ef51b5b 100644
--- a/src/history.rs
+++ b/src/history.rs
@@ -484,15 +484,10 @@ pub fn unapply_filter(
                 if tid != git2::Oid::zero() {
                     transaction.repo().find_tree(tid)?
                 } else {
-                    // This used to be our only fallback for the parent_count > 1 case.
-                    // It should never happen anymore.
-                    tracing::warn!("rejecting merge");
-                    let msg = format!(
-                        "rejecting merge with {} parents:\n{:?}",
-                        parent_count,
-                        module_commit.summary().unwrap_or_default()
-                    );
-                    return Err(josh_error(&msg));
+                    // This can happen when there is a merge commit some of whose parents are above the original target,
+                    // and some below. It should mean that the tree is identical for the filtered part, so we just have to
+                    // pick a side for the reconstructed unapplied history.
+                    transaction.repo().find_tree(new_trees[0])?
                 }
             }
         };
```
But I am a bit worried about potential side-effects here since I don't really understand what is happening.^^

christian-schilling:
Thanks for reporting and trying out that hack.
I will have a closer look at this as soon as I get to spend some time at a computer. So probably Monday.

RalfJung:
I am trying to reproduce this in a testcase, but how do I even set `-o base` when using josh-filter directly? Is there a way to have the same effect as a 'push to a branch that does not exist yet, with a given base'?

RalfJung:
Finally! I managed to reproduce this in a testcase. I added it in https://github.com/RalfJung/josh/commit/5b269ecb620f1191404e76bea447c7c6876bc156; locally then I just run `bash run-tests.sh tests/filter/subtree_prefix.t` and it prints
```
--- tests/filter/subtree_prefix.t
+++ tests/filter/subtree_prefix.t.err
@@ -148,3 +148,12 @@
   $ josh-filter -s :at_commit=$SUBTREE_TIP[:prefix=subtree]:/subtree refs/heads/master --update refs/heads/subtree-sync >/dev/null
   $ git merge subtree-sync --no-ff >/dev/null
   $ josh-filter -s :at_commit=$SUBTREE_TIP[:prefix=subtree]:/subtree refs/heads/master --update refs/heads/subtree --reverse
+  rejecting merge with 2 parents:
+  ab15d2dba8901016c749a7658ad42789b138f1e0
+  "Merge branch 'subtree-sync' into subtree"
+  original target: 066aefaaf53f87fd449c855f433de14e00c8187b
+  tree 1: 56508191b2958f7241e1b4944a207767989ebabf
+  tree 2: 3187811da10ce79a774749c54490c5edf2195ae1
+  [1] :prefix=subtree
+  [10] :/subtree
+  [10] :at_commit=c036f944faafb865e0585e4fa5e005afa0aeea3f[:prefix=subtree]
```
(that error contains a bunch of extra debug info that I added in josh locally)

christian-schilling:
So, I tried your test and fix. While the final result looks ok, there is an issue with the commit in question (that caused the error):
On that commit, "feature1" is missing, because it is only present on the second parent. While this might still work, it seems incorrect to me. The result we get when picking the second parent look correct however.
So I was thinking "what other criteria other than one being the first and the other the second do we have here to make a decision?"
And the answer I came up with: One is a result of filtering (it has a corresponding original commit in the main repo) while the other was authored in the partial repo. It seems to me that in that case it would be reasonable to choose the commit that *has* an original to supply the remainder of the tree. Like so:
```
if find_original(
    transaction,
    filterobj,
    original_target,
    original_parents_refs[0].id(),
    false,
)? == git2::Oid::zero()
{
     transaction.repo().find_tree(new_trees[1])?
} else {
     transaction.repo().find_tree(new_trees[0])?
}
```

Then we get:
```
$ git log --graph --pretty=%H:%s refs/heads/master
  *   6ac0ba56575859cfaacd5818084333e532ffc442:Merge branch 'subtree-sync' into subtree
  |\  
  | *   38a6d753c8b17b4c6721050befbccff012dfde85:Merge branch 'feature2'
  | |\  
  | | * 221f5ceab31209c3d3b16d5b2485ea54c465eca6:feature2
  * | | 75e90f7f1b54cc343f2f75dcdee33650654a52a6:subfeature2
  * | | 3fa497039e5b384cb44b704e6e96f52e0ae599c9:Merge branch 'subtree-sync' into subtree
  |\| | 
  | * |   2739fb8f0b3f6d5a264fb89ea20674fe34790321:Merge branch 'feature1'
  | |\ \  
  | | * | dbfaf5dd32fc39ce3c0ebe61864406bb7e2ad113:feature1
  | | |/  
  * | / 59b5c1623da3f89229c6dd36f8baf2e5868d0288:subfeature1
  |/ /  
  * | 103bfec17c47adbe70a95fca90caefb989b6cda6:add even more content
  * | 41130c5d66736545562212f820cdbfbb3d3779c4:subtree edit from main repo
  * | 0642c36d6b53f7e829531aed848e3ceff0762c64:subtree merge
  |\| 
  | * c036f944faafb865e0585e4fa5e005afa0aeea3f:add file2 (in subtree)
  * 0b4cf6c9efbbda1eada39fa9c1d21d2525b027bb:add file1

  $ git ls-tree --name-only -r 3fa497039e5b384cb44b704e6e96f52e0ae599c9
  feature1
  file1
  subtree/file2
  subtree/subfeature1
```

christian-schilling:
The good thing about this problem and possible side effect of possible fixes: They are on the "backward path" (only effect push) and that means we can still change the behavior in the future without breaking compatibility with existing repos.

RalfJung:
Yes that makes a lot of sense -- if a parent has feature1, it'd be strange for that to disappear and re-appear later.

So there could even be a sanity check here, where we *require* (at least) one of the commits to have an original, and then pick that one?

christian-schilling:
> So there could even be a sanity check here, where we _require_ (at least) one of the commits to have an original, and then pick that one?

We can certainly do that. Until we find the case where even that is not true and then need to fix that as well :)

I'm pretty sure by the way that this has nothing to do at all with the `at_commit` selective filtering.

RalfJung:
> I'm pretty sure by the way that this has nothing to do at all with the at_commit selective filtering.

Agreed. But after finding a testcase that reproduces this with at_commit I did not have the time/motivation to also build a testcase that avoids selective filtering. ;)

christian-schilling:
> Hm, at least for Miri this does not seem to help...

That's because I got it wrong at first and you copied my error...try this:
https://github.com/josh-project/josh/pull/994

RalfJung:
I opened an issue for the Miri trouble, to avoid spreading this across too many PRs: https://github.com/josh-project/josh/issues/998.

docs/src/reference/filters.md
```diff
@@ -91,6 +91,13 @@ These filter do not modify git trees, but instead only operate on the commit gra
 Produce a filtered history that does not contain any merge commits. This is done by
 simply dropping all parents except the first on every commit.
 
+### Filter specific part of the history **:at_commit=<sha>[:filter]**
```
LMG, line 94: out of date

LMG (approved):

PR #1011, opened 2022-11-09
SSH support, part 1 (superseded)

vlad-ivanov-name:
* add a custom shell for use with SSH
* add an RPC interface for running commands from the main process
* add an endpoint where git-upload-pack launch can be requested
* prepare a multi-service docker image
* add a userland ssh server (only for development and testing)


LMG:
🎉

PR #1012, opened 2022-11-10
SSH support, part 1

vlad-ivanov-name:
* add a custom shell for use with SSH
* add an RPC interface for running commands from the main process
* add an endpoint where git-upload-pack launch can be requested
* prepare a multi-service docker image
* add a userland ssh server (only for development and testing)

Change: If3ace8e012c5608d5f589589b7d4931dd6a9e185

---

**Stack**:
- #1019
- #1016
- #1015
- #1014
- #1013
- #1012 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*


josh-proxy/src/bin/josh-proxy.rs
```diff
@@ -565,35 +792,32 @@ async fn call_service(
...
+
+    if !josh_proxy::auth::check_auth(&remote_url, &auth, http_auth_required)
+        .in_current_span()
+        .await?
     {
         tracing::trace!("require-auth");
         let builder = Response::builder()
-            .header("WWW-Authenticate", "Basic realm=User Visible Realm")
+            .header(
+                reqwest::header::WWW_AUTHENTICATE,
```
christian-schilling, line 805: reqwest is a crate dependency that I'd like to get rid of, so I'd prefer not introducing additional usage of it's symbols.

christian-schilling (approved):

PR #1023, opened 2022-11-13
Prepare support for using SSH remotes during fetch

vlad-ivanov-name:
**Stack**:
- #1026
- #1023 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

### DIFF
--- a/josh-proxy/src/bin/josh-proxy.rs
+++ b/josh-proxy/src/bin/josh-proxy.rs
@@ -18,6 +18,7 @@ use indoc::formatdoc;
 use josh::{josh_error, JoshError};
 use josh_rpc::calls::RequestedCommand;
 use josh_rpc::tokio_fd::IntoAsyncFd;
+use serde::Serialize;
 use std::collections::HashMap;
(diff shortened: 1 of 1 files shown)


tests/proxy/get_remote.t
```diff
@@ -2,4 +2,4 @@
   $ cd ${TESTTMP}
 
   $ curl -s http://localhost:8002/remote
-  http://localhost:8001 (no-eol)
+  {"Http":"http://localhost:8001"} (no-eol)
```
christian-schilling, line 5: Will this break the UI?

christian-schilling: So this means UI does not work if there is no http remote? Makes some sense. How would it authenticate...
So for now, yes, error if there is none. As long as there is no concrete use case of exposing ssh remote (like there is with http) we don't need an API.

christian-schilling (approved):

PR #1040, opened 2022-11-16
Don't set #![deny(warnings)] on the whole crate

glittershark:
deny(warnings) is generally considered to be a pretty strong anti-pattern[0] because new warnings can be introduced, either by new versions of rustc or by new versions of dependencies, that the binary author may want to prevent but shouldn't prevent users from building from source. In this case, a newer version of `chrono` deprecated `NaiveDateTime::from_timestamp`, which ended up breaking my CI pipeline because josh is no longer able to build.

Generally, instead, it's better practice to deny *specific* warnings, or build with `-D warnings` in CI.

[0]: https://rust-unofficial.github.io/patterns/anti_patterns/deny-warnings.html

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,4 +1,3 @@
-#![deny(warnings)]
 #![warn(unused_extern_crates)]
 
 #[macro_export]


christian-schilling:
I ran into that problem myself a while ago. I still don't quite understand why a build breaking warning through a dependency change can happen with all dependencies pinned (including rustc) 🤔

But passing this from the outside via a flag does sound ok to me.

PR #1095, opened 2022-12-31
fix splash in site does not show.

MisileLab:
Splash in site does not show like
image
I changed url, so now can see splash! (I locally change url by browser devtool)
image


docs/src/intro.md
```diff
@@ -1,4 +1,4 @@
-![Just One Single History](../../splash.png)
+![Just One Single History](https://raw.githubusercontent.com/MisileLab/josh/master/splash.png)
```
christian-schilling, line 1: This now points to a different repo. We can't merge it like that.

christian-schilling (approved):

PR #1134, opened 2023-02-17
Don't serve all refs when explicit `@ref` is specified

vlad-ivanov-name:


christian-schilling:
> @christian-schilling this still serves all refs when we request `refs/...`, should that be changed?

yes, it should

josh-proxy/src/bin/josh-proxy.rs
```diff
@@ -792,15 +804,31 @@ fn is_repo_blocked(meta: &MetaConfig) -> bool {
     false
 }
 
-fn headref_or_default(headref: &str) -> String {
+#[derive(Clone, Debug)]
+enum HeadRef {
+    Explicit(String),
+    Implicit(String),
```
christian-schilling, line 810: Will this ever anything else than "HEAD" ?

christian-schilling (approved):

PR #1135, opened 2023-02-19
Start gitoxide integration

vlad-ivanov-name:


src/bin/josh-filter.rs
```diff
@@ -417,7 +420,8 @@ fn run_filter(args: Vec<String>) -> josh::JoshResult<i32> {
 
     if let Some(query) = args.get_one::<String>("query") {
         let repo = git2::Repository::open_from_env()?;
-        let transaction = josh::cache::Transaction::new(repo, None);
+        let oxide_repo = gix::ThreadSafeRepository::open(&env::current_dir()?)?;
```
christian-schilling, line 423: This should be `repo.path()` instead of `current_dir`. git2 might read this from an env variable and then those two handles have been initialized with different paths.

christian-schilling:
> @christian-schilling I'm not sure to which extent gitoxide should be integrated before this PR can be merged 🤔 lmk what you think

I think just putting the gix repo handle on Transaction and having it initialized is already enough for a first changeset.

christian-schilling (approved):

PR #1136, opened 2023-02-19
Return to change view when clicking commit message

christian-schilling:
Change: return-to-change


LMG:
The cursor doesn't change, you need to add `cursor: pointer`

LMG:
ok, let's ignore it since it's fixed in the next commit.

LMG (approved):

PR #1206, opened 2023-05-11
Fix download of git-lfs on some setups

LMG:
Change: lfs-fix


LMG:
This should not be needed. Investigating my setup.

PR #1208, opened 2023-05-11
Fix network access during build on some setups

LMG:
Change: network-fix


LMG:
This should not be needed. Investigating my setup.

PR #1213, opened 2023-05-29
Fix handlebars templates reporting missing objects 

christian-schilling:


LMG (commented):
Looks good, but what about making it a helper since all of the code is super similar? Also you need to run cargo fmt again (or maybe your version is wrong)

josh-core/src/query.rs
```diff
@@ -41,8 +54,26 @@ impl GraphQLHelper {
             variables.insert(k.to_string(), juniper::InputValue::scalar(v.render()));
         }
```
LMG, line 56: maybe a comment here to explain that one case is josh-filter and the other is josh-proxy?

josh-core/src/query.rs
```diff
@@ -162,13 +193,25 @@ pub fn render(
     drop(obj);
     drop(tree);
 
+    let repo_path =
+        if transaction.repo().path().file_name() == Some(std::ffi::OsStr::new("overlay")) {
```
LMG, line 197: can that be a problem if the repo is really called overlay? (I guess that will never be the case with our docker deployement)

christian-schilling: That's a good point 🤔 
This will be a problem when using the cli on a repo called "overlay" so not so good. I'll find a different way of doing this.

christian-schilling: Done

PR #1279, opened 2023-09-10
Use new "merge twice" algorithm to resolve push conflicts

christian-schilling:
Change: merge-twice

### DIFF
--- a/josh-core/src/history.rs
+++ b/josh-core/src/history.rs
@@ -461,23 +461,51 @@ pub fn unapply_filter(
 
                 if tid == git2::Oid::zero() && parent_count == 2 {
                     // If we could not select one of the parents, try to merge them.
+                    // We expect conflicts to occur only in the paths that are present in
+                    // the filtered commit.
+                    // As we are going to replace the contents of these files with commit being
+                    // pushed, we can ignore those conflicts. To do that we perform the merge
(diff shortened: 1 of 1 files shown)


josh-core/src/history.rs
```diff
@@ -461,23 +461,51 @@ pub fn unapply_filter(
...
                     // If we could not select one of the parents, try to merge them.
-
-                    if let Ok(mut merged_index) = transaction.repo().merge_commits(
+                    // We expect conflicts to occur only in the paths that are present in
+                    // the filtered commit.
+                    // As we are going to replace the contents of these files with commit being
+                    // pushed, we can ignore those conflicts. To do that we perform the merge
+                    // twice: Once with the "ours" and once with the "theirs" merge file favor.
+                    // After that we do "unapply()" on both resulting trees, which will replace
+                    // The files selected by the filter with the content being pushed.
```
RalfJung, line 470:
```suggestion
                    // the files selected by the filter with the content being pushed.
```

josh-core/src/history.rs
```diff
@@ -461,23 +461,51 @@ pub fn unapply_filter(
...
+                    // If our assumption was correct and all conflicts where in filtered files
```
RalfJung, line 471:
```suggestion
                    // If our assumption was correct and all conflicts were in filtered files
```

PR #1285, opened 2023-10-09
josh-proxy.rs: bind dualstacked

flokli:
Binding on 0.0.0.0 will cause josh to only listen on IPv4, while specifying `[::]` will bind on *both* IPv4 and IPv6.

With this change:

```
❯ ss -ltpn | grep 8000
LISTEN 0      128                *:8000            *:*    users:(("josh-proxy",pid=1894190,fd=10))
```


flokli:
Thanks!

PR #1303, opened 2023-12-23
Optionally create branches on most recent similar commit

LMG:


christian-schilling (changes requested):
The code looks reasonable so far. But how should the functionality be exposed? It also needs a test.

PR #1304, opened 2023-12-26
Refactor and annotate functions in history handling

vlad-ivanov-name:
This PR doesn't change any behaviour, but rather adds comments and changes variable names in order for code to be more readable -- I hope that comments are correct

### DIFF
--- a/josh-core/src/filter/mod.rs
+++ b/josh-core/src/filter/mod.rs
@@ -833,12 +833,18 @@ pub fn unapply<'a>(
     }
 
     if let Op::Chain(a, b) = to_op(filter) {
-        let i = if let Ok(i) = invert(a) { invert(i)? } else { a };
-        let p = apply(transaction, i, parent_tree.clone())?;
+        // If filter "a" is invertable, use "invert(invert(a))" version of it, otherwise use as is
+        let a_normalized = if let Ok(a_inverted) = invert(a) {
+            invert(a_inverted)?
+        } else {
+            a
+        };
+        let filtered_parent_tree = apply(transaction, a_normalized, parent_tree.clone())?;
+
         return unapply(
             transaction,
             a,
-            unapply(transaction, b, tree, p)?,
+            unapply(transaction, b, tree, filtered_parent_tree)?,
             parent_tree,
         );
     }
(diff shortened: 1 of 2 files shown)


josh-core/src/filter/mod.rs
```diff
@@ -824,12 +824,18 @@ pub fn unapply<'a>(
     }
 
     if let Op::Chain(a, b) = to_op(filter) {
-        let i = if let Ok(i) = invert(a) { invert(i)? } else { a };
-        let p = apply(transaction, i, parent_tree.clone())?;
+        // If filter "a" is invertable, use "invert(invert(a))" version of it, otherwise use as is
```
christian-schilling, line 827: This comment states the obvious but does not explain why the double invert is done. So it's a good this that you poke at this because it makes me realize that I don't remember why this is done and should have commented back when I wrote it... I'll investigate.

josh-core/src/history.rs
```diff
@@ -75,32 +75,44 @@ pub fn walk2(
 
 fn find_unapply_base(
     transaction: &cache::Transaction,
-    bm: &mut HashMap<git2::Oid, git2::Oid>,
+    // Used as a cache to avoid re-applying the filter to the same commit -
+    // this function is called during revwalk so there be a lot of repeated
+    // calls
+    filtered_to_original: &mut HashMap<git2::Oid, git2::Oid>,
     filter: filter::Filter,
+    // "contained in" refers to OID of the commit in the original (unfiltered) history
```
christian-schilling, line 83: That's not what `contained_in` means. What it does mean: When building the `bm`/`filtered_to_original` mapping use this as a starting point for the search for orginals. Meaning if there are multiple originals that map to the same filtered commit (which is common) use one that is reachable from `contained_in`. Or in other words one that is contained in the the history of `contained_in`.

josh-core/src/history.rs
```diff
@@ -256,9 +272,12 @@ fn find_new_branch_base(
     };
     tracing::info!("new branch base?");
 
+    // Walk filtered history tree, trying to find a base for every commit
```
christian-schilling, line 275: remove the word `tree` -> It has different meaning in git that is misleading here. The sentence works without it.

christian-schilling (changes requested):
See comments

PR #1309, opened 2024-01-12
Support cross workspace amend

christian-schilling:
Change: ws-amend


LMG:
This is quite old and would be done differently today, especially in the light of the josh CLI tool

PR #1323, opened 2024-04-04
Support push + force push over SSH

vlad-ivanov-name:
* Support receive-pack over SSH
* Improve tracing in push routines
* Improve error propagation between hooks and proxy
* Improve push option handling: use predefined struct
* Add `force` push option

commit-id:2dbf5e93

### DIFF
--- a/josh-proxy/src/bin/josh-proxy.rs
+++ b/josh-proxy/src/bin/josh-proxy.rs
@@ -22,9 +22,10 @@ use josh::{josh_error, JoshError, JoshResult};
 use josh_rpc::calls::RequestedCommand;
 use serde::Serialize;
 use std::collections::HashMap;
+use std::ffi::OsStr;
 use std::io;
 use std::net::IpAddr;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::process::Stdio;
 use std::str::FromStr;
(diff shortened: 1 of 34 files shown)


josh-proxy/src/bin/josh-proxy.rs
```diff
@@ -81,6 +82,28 @@ impl JoshProxyUpstream {
     }
 }
 
+enum GitEnvVar {
```
christian-schilling, line 85: I don't see the advantage of introducing this type over just using the string constants in place. What's the reasoning?

christian-schilling (commented):
Change looks fine to me apart from the question about `GitEnvVar`.

josh-proxy/src/bin/josh-proxy.rs
christian-schilling: I'd prefer just keeping the strings in place like they were. Typos don't seem like a real threat to me. And having an extra thing to look up to see what something is just hurts readability. Even more so with the extra complication of going through a trait for conversion.
Also now there is a mix of some strings being inline constants and some being in this enum. Even just among the git env vars. Strings are simple. Let's just keep it simple.

christian-schilling (approved):

PR #1324, opened 2024-04-08
Fix percent encoded parameters

christian-schilling:
Change: percent-decode


christian-schilling:
That was my first attempt. I already changed all the code to use Url form the `url` create (had to move things around a bit for that) to discover that 1) `request.uri()` from hyper looks like a relative url and will make `url::Url` panic.
2) when used from `josh-filter` there is no URL at all so this also does not work.
So I thought writing it is not so much hassle compared to that.
I did not think about using `form_urlencoded`. First thing that came up when trying that one was that `josh-core` does not already depend on `url`. And then that `form_urlencoded` was removed from the `url` create some time ago.
Now there is a dedicated create to just `form_urlencoded`...So I added that as a dep to `josh-core`...

PR #1331, opened 2024-05-14
Fix race condition in check_auth

vlad-ivanov-name:
When a lot of requests arrive roughly at the same time, several requests can enter the critical section where an HTTP request to upstream is made to check the auth provided by the client. This means that potentially thousands of requests can get through to the remote, leading to rate limits and network errors with some remotes.

* Fix the issue by extending the lock region
* Switch mutex to async to avoid blocking runtime
* Improve tracing

commit-id:d44447ef

### DIFF
--- a/josh-proxy/src/auth.rs
+++ b/josh-proxy/src/auth.rs
@@ -69,8 +69,8 @@ impl Handle {
                 .ok_or_else(|| josh::josh_error("no auth found"))?;
 
             let line = String::from_utf8(line)?;
-            let token = line
-                .strip_prefix("Basic ")
+            let (_, token) = line
+                .split_once(' ')
                 .ok_or_else(|| josh::josh_error("Unsupported auth type"))?;
 
             let decoded = BASE64.decode(token)?;


josh-proxy/src/auth.rs
```diff
@@ -74,20 +74,34 @@ pub fn add_auth(token: &str) -> josh::JoshResult<Handle> {
...
 
 pub async fn check_auth(url: &str, auth: &Handle, required: bool) -> josh::JoshResult<bool> {
+    use opentelemetry_semantic_conventions::trace::HTTP_RESPONSE_STATUS_CODE;
+
     if required && auth.hash.is_empty() {
         return Ok(false);
     }
 
-    if let Some(last) = AUTH_TIMERS.lock()?.get(&(url.to_string(), auth.clone())) {
+    let mut auth_timers = AUTH_TIMERS.lock().await;
```
christian-schilling, line 83: With the lock now spanning the whole function including the http request itself, I see a potential problem when multiple requests with different users arrive at the same time. Those will now be completely serial.
Might not be a big issue though.

christian-schilling (approved):

Issue #1345, opened 2024-08-01
Non-roundtrip push to josh branch when a commit has a `.` as the last character of the author name


RalfJung:
The commit https://github.com/rust-lang/rust-analyzer/commit/5af51fa47610d0f88dcab99aaaff2ee4fca17c00 does not properly round-trip through a push to josh.

Original commit info (`git show --format=raw`):
```
commit 5af51fa47610d0f88dcab99aaaff2ee4fca17c00
tree 5cf661188e20157ffab5d742ce22d5fa44d18bd1
parent 4a91453409aaf58eba0a91e4b62a35f49814b394
author Winston H. <56998716+winstxnhdw@users.noreply.github.com> 1721814140 +0100
committer GitHub <noreply@github.com> 1721814140 +0100
```
After a round-trip through josh:
```
commit fc919994d0759bca7423444e0d0bc044df80359a
tree 5cf661188e20157ffab5d742ce22d5fa44d18bd1
parent 4a91453409aaf58eba0a91e4b62a35f49814b394
author Winston H <56998716+winstxnhdw@users.noreply.github.com> 1721814140 +0100
committer GitHub <noreply@github.com> 1721814140 +0100
```
The `.` after `Winston H` went missing. Seems like somewhere josh is normalizing usernames and doesn't like the trailing `.`?

RalfJung:
Seems like git [used to](https://stackoverflow.com/questions/26159274/is-it-possible-to-have-a-trailing-period-in-user-name-in-git) prevent a `.` from even existing at the end of a username, but this got [changed last year](https://github.com/git/git/commit/1c04cb0744d2acdcaebc77b0e78c47efbba67fd5). So probably whatever git library josh is using needs to get the same kind of update?

RalfJung:
This change was mirrored in libgit2 via https://github.com/libgit2/libgit2/commit/cf320cc125e589a4de109de9098d54098f03b31b, which was included in libgit2 1.8. That needs [version 0.19](https://github.com/rust-lang/git2-rs/blob/master/CHANGELOG.md#0190---2024-06-13) of the git2 Rust crate.

PR #1346, opened 2024-08-01
bump git2 dependency to 0.19

RalfJung:
This fixes https://github.com/josh-project/josh/issues/1345.

Does it need a cache version bump? In my tests, this PR suffices to fix the issue even with an existing cache that was used for the previous "bad" conversion.


lnicola (approved):
I can confirm this fixes the RA issues.

Might be worth including a `cargo update -p time` to fix a build issue on recent compilers, but that's unrelated.

christian-schilling (changes requested):
This will need a bump of `CACHE_VERSION` in `cache.rs`. In the case of this bug manifesting the resulting sha's will be different between versions and thus the cache cannot be kept from previous versions.

christian-schilling (approved):

PR #1387, opened 2024-09-14
Fix round-tripping when commit contains non-standard headers

### LINKED ISSUE #1377: Roundtripping broken when commit includes custom headers (e.g. GitButler)
In https://github.com/josh-project/josh/issues/596#issuecomment-2298514080, I discovered the reason my SHAs were not lining up, by finding a specific commit where the histories began to diverge (between original repo and imported-then-subpath-cloned repo).

In this case, it seems that the [GitButler](https://gitbutler.com/) tool is injecting custom headers into the commit object, which are being elided when JOSH rewrites them:

```sh-session
❯ diff <(git show --pretty=raw 8f49df1dc) <(git show --pretty=raw 34d061bdc)
1c1
< commit 8f49df1dc0acf9389c7aa712847cc8309f01755d
---
> commit 34d061bdca18e243ccc8bfe53333037ce17500f9
6,7d5
< gitbutler-headers-version 2
< gitbutler-change-id baefc019-8ef0-40f2-9f1f-dc92b9984387
```

Is it possible to support unknown headers or do new headers need to be "learned" case-by-case?

Would it be possible to update JOSH to support these headers? I don't even know what GitButler would do if the `gitbutler-change-id` was the same UUID for a different commit object, though. Maybe @schacon would be able to clarify if this would be a problem...

bjeanes:
Fixes #1377
Closes #1389

This takes a similar strategy as GitButler uses to insert its custom headers in the first place. In essence, we build the bytes of the commit object ourselves, starting with the bytes from the original commit, and then write it using `git2::odb::write`.

The only thing I am a little unhappy with is how committer/author handling has to "know" the Git format for timestamps. Without using `unsafe {}`, I'm not sure how to leverage `libgit2`'s authoritative implementation for these parts in a clean way, and even with `unsafe {}` I'm not sure it's possible. I _think_ this is good enough though, but I am interested in others' thoughts.

NB: I am familiar enough with Rust to have made this work, but I am far from a "good" Rust programmer, so I am happy to take direction to tweak this implementation if you feel there are specific improvements (e.g. such as standardising `BString`/`String`/`&str`).


bjeanes:
Some further thoughts:

- Access to custom headers could allow the `unsign` filter to be reversible, by merely _renaming_ the `gpgsig` (which it seems can be done out-of-the-box with git2 [by passing `commit_signed` a `signature_field: Some("orig-gpgsig")](https://docs.rs/git2/latest/git2/struct.Repository.html#method.commit_signed)`
- The PR as implemented might be possible with `gix-odb` and `gix-object` instead of `git2` (I note that `gix` is already a dependency). Its [`CommitRef`](https://docs.rs/gix-object/0.44.0/gix_object/struct.CommitRef.html) provides `mut` access to its fields and it already [exposes `extra_headers`](https://docs.rs/gix-object/0.44.0/gix_object/struct.CommitRef.html#structfield.extra_headers). The object can be loaded with [`gix_odb::store::Handle::try_find()`](https://docs.rs/gix-odb/0.63.0/gix_odb/store/struct.Handle.html#method.try_find) and written with [its `write()`](https://docs.rs/gix-odb/0.63.0/gix_odb/store/struct.Handle.html#method.write) (`CommitRef` impls the requisite `WriteTo` interface). `gix` is not used _much_ in JOSH, so I'll leave this as-is _for now_, but using a well-maintained API for this might be preferable.

Looking forward to the thoughts of maintainers and happy to throw up a version using `gix` (if I'm successful) if there's interest. Also happy to merge this and follow up with that work or to make minor changes to what I have already done, as I'm pretty motivated to help get JOSH working for the repository I want to import into monorepo.

bjeanes:
Thanks for unblocking the tests. The only failure seems to be:

```
   $ curl -s http://localhost:8002/version
-  Version: r*.*.* (glob)
+  Version: v24.08.14-5-g5a26dd2
```

I didn't change the version format, so I'm not sure why it's failing.

LMG:
Yes, I saw that. This is very curious. I don't have a lot of time to look at it right now though, and I also would like a review from @christian-schilling  before merging

bjeanes:
I just realised that it should be possible to use [`gix_object::Commit`](https://docs.rs/gix-object/0.44.0/gix_object/struct.CommitRef.html)/[`gix_object::CommitRef`](https://docs.rs/gix-object/0.44.0/gix_object/struct.CommitRef.html) without using the rest of Gix or introducing the IO concerns raised in #1337. This would avoid needing to introduce this bespoke `CommitBuffer` and keep the implementation of custom header handling to a more purpose built location.

I'm going to have a little look-see if I can make an improved version.

bjeanes:
> Yes, I saw that. This is very curious. I don't have a lot of time to look at it right now

It looks like it's just because the most recent release tag changed schemes (possibly unintentionally):

```sh-session
❯ git tag                                       
r21.03.07
r21.03.08
r21.03.09
r21.03.19
r21.03.19.1
r21.05.05
r21.08.16
r21.09.15
r21.11.21
r22.03.17
r22.03.18
r22.03.24
r22.04.15
r22.06.08
r22.06.14
r22.06.22
r22.08.25
r22.09.02
r22.09.27
r22.12.03
r22.12.06
r23.02.14
r23.12.04
v0.1.0
v0.2.0
v24.08.14       <---
```

bjeanes:
Closing in lieu of #1389

PR #1389, opened 2024-09-17
Fix round-tripping commits that contain extra headers

### LINKED ISSUE #1377: Roundtripping broken when commit includes custom headers (e.g. GitButler)

bjeanes:
Fixes #1377
Closes #1387

As per https://github.com/josh-project/josh/pull/1387#issuecomment-2355037601 this is how this might look by using the `gix-object` crate to own the commit object format.


josh-core/src/history.rs
```diff
@@ -196,49 +196,70 @@ pub struct RewriteData<'a> {
...
-        .unwrap_or(base.message_raw().unwrap_or("no message").to_string());
-    let tree = &rewrite_data.tree;
-
-    let a = base.author();
-    let new_a = if let Some((author, email)) = rewrite_data.author {
-        git2::Signature::new(&author, &email, &a.when())?
-    } else {
-        a
-    };
+    use bstr::ByteSlice;
```
christian-schilling, line 208: I'd prefer to not have `use` but instead fully qualify the identifiers in place.

josh-core/src/history.rs
```diff
@@ -196,49 +196,70 @@ pub struct RewriteData<'a> {
...
+    }
+
+    commit
+        .extra_headers
+        .retain(|(k, _)| k.as_bytes() != "gpgsig".as_bytes());
+
+    let mut b = vec![];
+    commit.write_to(&mut b)?;
 
     if let (false, Ok((sig, _))) = (unsign, repo.extract_signature(&base.id(), None)) {
```
christian-schilling, line 253: Would't it be possible to combine the the handling of the signature with the remaining headers. That is,
` commit
         .extra_headers
         .retain(|(k, _)| (k.as_bytes() != "gpgsig".as_bytes()) || !unsign);`

and getting rid of the extra `commit_signed` below?

I think it might even necessary to do so depending on how headers ordering is handled/specified (or rather unspecified)

tests/proxy/roundtrip_custom_header.t
```diff
@@ -0,0 +1,182 @@
+  $ . ${TESTDIR}/setup_test_env.sh
```
christian-schilling, line 1: The test case for this should be written using `josh-filter` and not using the proxy. Those are easier and faster to run. Proxy tests should only be used for behaviour that is unique to the proxy.

josh-core/src/history.rs
bjeanes: I'll give that a try and see if the tests pass!

tests/proxy/roundtrip_custom_header.t
bjeanes: Yeah I had wondered about this. I started here because that's how I initially discovered the bug. I do actually have the filter test already written but I was a bit confused by how `--update` and `--reverse` interact and wasn't confident that it was properly reproducing the issue.

I'll swap the tests and then when you re-review perhaps you can verify the approach.

bjeanes:
> I think it's ok to use gix here.

Yeah once I realised (after doing #1387) that it was possible to use `gix` without using gix's odb or repo, then this was my strong preference for approach. I'm glad you agree. I'll close #1387.

What do you want to do about the `/version` endpoint and test? I changed the test but I'm not sure if the latest git tag on the repo was meant to be `r24.08.14` instead of `v24.08.14`, to match the previous scheme.

josh-core/src/history.rs
bjeanes: Hmm... this is a trait not a module, though...

In any case, it looks like this is somewhat vestigial from an earlier approach. The only place this is needed can be replaced with a deref. This is done in 053d5e35ca7d48173b32723064eac3bef53fc48a

I'm interpreting this feedback to also apply to `use gix_object::{CommitRef, WriteTo};` but have broken it out as a separate commit 31ab08d43d837e9ac09b96c96379b33e5031ce02 in case not.

tests/proxy/roundtrip_custom_header.t
bjeanes: I've replaced the test with a filter test but it currently is broken because of my confusion about the above. If you could give some guidance that'd be excellent.

bjeanes: Never mind! I took another approach by applying two different but inverse filters instead of using `--reverse`. `:prefix=foo` followed by `:/foo` should actually be closer to what the proxy test was doing anyway.

josh-core/src/history.rs
```diff
@@ -205,40 +205,49 @@ pub fn rewrite_commit(
...
+        .map(|x| format!("{}", x.id()))
+        .collect::<Vec<_>>();
 
-    let c = base.committer();
-    let new_c = if let Some((committer, email)) = rewrite_data.committer {
-        git2::Signature::new(&committer, &email, &c.when())?
-    } else {
-        c
-    };
+    let mut commit = gix_object::CommitRef::from_bytes(odb_commit.data())?;
```
Byron, line 220: The idea here would be to call `.to_owned()` so dealing with changing the commit is a little less cumbersome, hopefully.

You might also find [this method](https://github.com/Byron/gitoxide/blob/abcfb659786425ec09eff6b644cd2ad36b7d6bc4/gix-object/src/commit/ref_iter.rs#L57), as it's related to handling signatures, even though I am not sure it perfectly applies here.

bjeanes: Thanks for stopping by @Byron!

> The idea here would be to call `.to_owned()` so dealing with changing the commit is a little less cumbersome, hopefully.

I had wondered about this. I opted to stay with the ref struct once I had it working but agree the Owned version might be easier. I might do that and compare, but I'm pretty happy with how it is now.

The thing I found the most cumbersome was not the lifetimes but he handling of IDs as hex, actually. Because it was a byte slice and taking bytes, it assumed this would be 20 raw bytes, as per `git2` and was surprised that this was 40-char hex bytes. I think being able to more naturally work with both would be an ergonomic improvement.

> You might also find [this method](https://github.com/Byron/gitoxide/blob/abcfb659786425ec09eff6b644cd2ad36b7d6bc4/gix-object/src/commit/ref_iter.rs#L57), as it's related to handling signatures, even though I am not sure it perfectly applies here.

Because JOSH is rewriting commits in ways that will break the signature, the choice indicated by `unsign` is merely whether to remove the signature altogether or to leave it broken (so that the commit can be inversely rewritten). Validating it or using the signature itself doesn't seem applicable (in this area).

But that method does look convenient, so making a mental note!

Byron: You are welcome!

> The thing I found the most cumbersome was not the lifetimes but he handling of IDs as hex, actually. Because it was a byte slice and taking bytes, it assumed this would be 20 raw bytes, as per `git2` and was surprised that this was 40-char hex bytes. I think being able to more naturally work with both would be an ergonomic improvement.

The refs represent what's in the buffer, and the buffer contains hex-hashes. The fully owned version is at liberty to more naturally represent the data, so it would use `gix_hash::ObjectId` instead.

> But that method does look convenient, so making a mental note!

I think with the signature range (i.e. where the signature is in the bytes buffer) one could easily build an unsigned version of that buffer. That one could parse and alter, while leaving the signature out of there which I thought was happening here.

Admittedly I didn't look closely enough to truly understand what's happening, so this is merely a dump of what I know and you can mix and match :).

christian-schilling (approved):
Looks good now. Thanks a lot!

bjeanes:
Thank you @christian-schilling! What's your approach for release schedule. Now that this is fixed, a release including the fix will unblock my team's adoption of JOSH. Is this something that could be done?

christian-schilling:
> Thank you @christian-schilling! What's your approach for release schedule. Now that this is fixed, a release including the fix will unblock my team's adoption of JOSH. Is this something that could be done?

Out approach to release schedule is to release from time to time when major bugs of features where fixed or introduced. But we do run production instances straigt of `master`.
I did tag a new release.

Issue #1429, opened 2025-08-22
shell.nix outdated and failed on macOS: `darwin.apple_sdk` has been phased out.


Binlogo:
### Description

When attempting to build a Nix environment using ⁠nix-shell, the process fails with an error indicating that ⁠darwin.apple_sdk_11_0 has been removed. This appears to be due to a dependency in ⁠shell.nix that relies on this legacy SDK, which is no longer available in recent versions of Nixpkgs.

### Steps to Reproduce
1. Use a ⁠shell.nix file that depends on `⁠darwin.apple_sdk`.
2. Run ⁠nix-shell shell.nix on macOS, the build process will faild.

```
  System:
    OS: macOS 15.6
    CPU: (24) arm64 Apple M2 Ultra
    Memory: 627.47 MB / 64.00 GB
    Shell: 5.9 - /bin/zsh
```

The command fails with the following error message:
```
❯ nix-shell shell.nix         
error:
       … while calling the 'derivationStrict' builtin
         at <nix/derivation-internal.nix>:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'nix-shell'
         whose name attribute is located at /nix/store/gjl3hj01mbjmfdcydcwbldh1jh6c53h8-nixos-21.11/pkgs/stdenv/generic/make-derivation.nix:205:7

       … while evaluating attribute 'buildInputs' of derivation 'nix-shell'
         at /nix/store/gjl3hj01mbjmfdcydcwbldh1jh6c53h8-nixos-21.11/pkgs/stdenv/generic/make-derivation.nix:247:7:
          246|       depsHostHost                = lib.elemAt (lib.elemAt dependencies 1) 0;
          247|       buildInputs                 = lib.elemAt (lib.elemAt dependencies 1) 1;
             |       ^
          248|       depsTargetTarget            = lib.elemAt (lib.elemAt dependencies 2) 0;

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: darwin.apple_sdk_11_0 has been removed as it was a legacy compatibility stub; see <https://nixos.org/manual/nixpkgs/stable/#sec-darwin-legacy-frameworks> for migration instructions
```

### Possible Solution
[...]

PR #1430, opened 2025-08-22
Update shell.nix to use NixOS 24.11 and update `cram` to `prsyk`

Binlogo:
Fix shell.nix for macOS compatibility and update to NixOS 24.11
- Update from NixOS 25.05 (non-existent) to stable NixOS 24.11
- Fix macOS build failure by properly handling Darwin frameworks
  without using deprecated darwin.apple_sdk
- Add Security, CoreFoundation, and SystemConfiguration frameworks
  for macOS compatibility
- Simplify Python environment setup for prysk testing, pin to specific
version
- Remove virtual environment in favor of user-space pip installation

Fixes https://github.com/josh-project/josh/issues/1429

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,49 +1,58 @@
 let
   pkgs = import (fetchTarball {
-    name = "nixos-25.05";
-    url = "https://github.com/NixOS/nixpkgs/archive/refs/tags/25.05.tar.gz";
-    sha256 = "1915r28xc4znrh2vf4rrjnxldw2imysz819gzhk9qlrkqanmfsxd";
+    name = "nixos-24.11";
+    url = "https://github.com/NixOS/nixpkgs/archive/refs/tags/24.11.tar.gz";
+    # Hash obtained using `nix-prefetch-url --unpack <url>`
+    sha256 = "1250a3g9g4id46h9ysy5xfqnjf0yb2mfai366pyj5y2bzb8x0i2l";
   }) {};
   
-  pythonWithPryskDeps = pkgs.python3.withPackages (ps: with ps; [
+  # Handle darwin-specific dependencies properly
+  extra_deps = if pkgs.stdenv.isDarwin then [
+    pkgs.darwin.apple_sdk.frameworks.Security
+    pkgs.darwin.apple_sdk.frameworks.CoreFoundation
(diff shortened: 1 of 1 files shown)


Binlogo:
@vlad-ivanov-name Would you mind reviewing this pull request? Thanks!

vlad-ivanov-name:
Will review pending PRs this weekend, thanks for the ping.

shell.nix
```diff
@@ -22,10 +20,30 @@ in
...
+         echo "Creating Python virtual environment..."
+         python3 -m venv .venv
+       fi
+       
+       source .venv/bin/activate
+       
+       # Install or upgrade prysk
+       echo "Installing/updating prysk..."
+       pip install --upgrade pip
+       pip install --upgrade prysk
```
vlad-ivanov-name: Can we pin the version somehow? Ideally with a lockfile, and ideally with `uv`

shell.nix
```diff
@@ -22,10 +20,30 @@ in
...
+     ];
      RUST_BACKTRACE = 1;
+     
+     shellHook = ''
+       echo "Welcome to Josh development environment!"
+       echo "Rust version: $(rustc --version)"
+       echo "Cargo version: $(cargo --version)"
+       
+       # Install prysk using pip in a virtual environment
+       if [ ! -d ".venv" ]; then
```
vlad-ivanov-name: Isn't it the job of nix to maintain the environment? Do we even need the venv

shell.nix
vlad-ivanov-name: Looking at the diff this pinning used to be handled by nix itself so unless something radically changed in nix and we can't do it anymore, I'd keep it that way

Binlogo: Sure, I forgot to mention that the `prysk` package is missing in the `nixpkgs`. I've tried to add it upstream but failed due to dependency issues. I'll try again as I become more familiar with `nixpkgs`. So, here's a workaround.

The `uv` package is an inspiring one to use here, and I will update it.

vlad-ivanov-name: I see okay. So maybe then a uv setup with lockfile + a nix "layer" where uv installs prysk could work. I know we can't access internet normally in build step but it's possible with fixed output derivation

PR #1450, opened 2025-08-25
chore: remove jaeger

reneleonhardt:
### Chores
- Remove opentelemetry-jaeger
- Changed priority: 1. JOSH_OTLP_ENDPOINT 2. JOSH_JAEGER_ENDPOINT
- Update dependencies


reneleonhardt:
Please have a look if this is what you wanted, or even better try yourself locally:
https://github.com/open-telemetry/opentelemetry-rust/blob/main/docs/migration_0.28.md

vlad-ivanov-name:
we need to preserve grpc export as it was supported before, and it's the primary way of sending traces.

i've updated the code, it still needs two changes: 

1) unfortunately there seems to be a regression: when OTLP export is enabled, clone from josh stalls and seems to just hang; this needs some investigation, possibly with tokio-console to see if runtime is blocked.
2) service name needs to be set to `josh-proxy`

reneleonhardt:
Sorry about the service name, the api and docs looked like tracer_provider.tracer() could be it at first glance.

reneleonhardt:
Could the new block_on() be waiting for the new tracer_provider.shutdown()?
Can you comment the shutdown out and retry?
If that doesn't stall anymore you could reduce the default timeout of
```rust
.shutdown_with_timeout(Duration::from_secs(5))
```

vlad-ivanov-name:
>Could the new block_on() be waiting for the new tracer_provider.shutdown()?

it shouldn't normally even get to that point; run_proxy doesn't return until josh exits.

also, block_on is just expansion of tokio's `#[main]` macro, so it was there before

vlad-ivanov-name:
debugged this and it's about the exporter used -- we need to use batching exporter, the simple one is, well, for simple cases only, not really for production. simple exporter tries to submit the spans using `futures_executor::block_on` and i guess that fails while in a tokio thread pool.

with updated exporter i can now see traces in jaeger.

image

vlad-ivanov-name:
also verified env filter works.

reneleonhardt:
Great that you could find the batch exporter!

Maybe you can update the Makefile to make lokal debugging easier, start jaeger then start the josh export? 😄

vlad-ivanov-name:
I think docker-compose is the usual way of handling this.

PR #1469, opened 2025-09-08
Fix prefix_sort ordering

LMG:
prefix_sort was using a sorting function that required, since we updated rust, a total ordering.
In fact prefix_sort should reorder alphabetically, while preserving the order of filters that share a prefix as input or output directory.

This commits replaces the function so that it does so, preventing crashes in cases where the ordering was revealed to not be total.

### DIFF
--- a/josh-core/src/filter/opt.rs
+++ b/josh-core/src/filter/opt.rs
@@ -4,6 +4,9 @@
  */
 
 use super::*;
+use std::cmp::Ordering;
+use std::collections::HashMap;
+use std::collections::VecDeque;
 
 lazy_static! {
     static ref OPTIMIZED: std::sync::Mutex<std::collections::HashMap<Filter, Filter>> =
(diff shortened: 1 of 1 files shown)


LMG:
Here is a minimal reproducing example of the crash with the current version:
 ```
$ cat > f << EOF
  >  ::re.md
  > de = :/de:[
  >     ::b/
  > ]
  > depe = :/depe:[
  >     ::re.md
  >     ::b/
  > ]
  > do = :/do:[
  >     ::b/
  > ]
  > vi = :/vi:[
  >     ::b/
  > ]
  > g2jc = :/g2jc:[
  >     ::b/
  > ]
  > ge = :/ge:[
  >     ::ex/
  >     ::hc/
  > ]
  > ge-to-gh = :/ge-to-gh:[
  >     ::b/
  > ]
  > gi-sy = :/gi-sy:[
  >     ::b/
  > ]
  > go = :/go:[
  >     ::b/
  > ]
  > h2w = :/h2w:[
  >     ::b/
  > ]
  > it = :/it:[
  >     ::b/
  > ]
  > ja-tr = :/ja-tr:[
  >     ::b/
  > ]
  > je = :/je:[
  >     ::ad/
  >     ::ma/
  > ]
  > kp-sy = :/kp-sy:[
  >     ::re.md
  >     ::b/
  > ]
  > op = :/op:[
  >     ::b/
  > ]
  > pr = :/pr:[
  >     ::b/
  > ]
  > ra = :/ra:[
  >     ::b/
  >     ::po-sa/tr/b/
  >     ::po-sa/lo/b/
  > ]
  > sc = :/sc:[
  >     ::b/
  > ]
  > so = :/so:[
  >     ::so-ma/b/
  > ]
  > un-se = :/un-se:[
  >     ::b/
  > ]
  > EOF

  $ josh-filter -p --file f
```

I am not adding this to the test suite, as it is dependant on the specifics of the implementation of the current sorting function, which we do not want to test. In fact, this should be adequately covered by the current pretty_print.t.

vlad-ivanov-name:
this fixes https://github.com/josh-project/josh/issues/1413 right?

LMG:
Yes, I didn't notice this, we had the exact same problem

PR #1483, opened 2025-09-19
Add :prune=trivial-merge filter

christian-schilling:
The default history simplification will not drop merge commits if their tree differs from any parent.
In most cases however merges that don't differ from the first parent are not wanted in the output history.
This new filter allows to prune them.

Change: prune-trivial-merges

### DIFF
--- a/josh-core/src/filter/parse.rs
+++ b/josh-core/src/filter/parse.rs
@@ -35,6 +35,22 @@ fn make_op(args: &[&str]) -> JoshResult<Op> {
         ["SQUASH"] => Ok(Op::Squash(None)),
         ["SQUASH", _ids @ ..] => Err(josh_error("SQUASH with ids can't be parsed")),
         ["linear"] => Ok(Op::Linear),
+        ["prune", "trivial-merge"] => Ok(Op::Prune),
+        ["prune"] => Err(josh_error(indoc!(
+            r#"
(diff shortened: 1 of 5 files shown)


josh-core/src/filter/parse.rs
```diff
@@ -35,6 +35,7 @@ fn make_op(args: &[&str]) -> JoshResult<Op> {
         ["SQUASH"] => Ok(Op::Squash(None)),
         ["SQUASH", _ids @ ..] => Err(josh_error("SQUASH with ids can't be parsed")),
         ["linear"] => Ok(Op::Linear),
+        ["prune", "trivial-merge"] => Ok(Op::Prune),
```
vlad-ivanov-name, line 38: would help to improve error message here by handling "prune" without args (similar to workspace above) and "prune" with wrong args

josh-core/src/filter/parse.rs
```diff
@@ -35,6 +35,22 @@ fn make_op(args: &[&str]) -> JoshResult<Op> {
...
+
+            Note: use "=" to provide the argument value:
+
+              :prune=trivial-merge
+            "#
+        ))),
+        ["prune", _] => Err(josh_error(indoc!(
+            r#"
+            Filter ":prune" only supports "trivial-merge"
+            as arguement value.
```
LMG: argument

PR #1494, opened 2025-09-22
Add computed filter resolver support

friel-openai:
Add plumbing for computed filters

Introduce a ComputedFilter hook on transactions so a `:computed` filter can be resolved per commit, cached, and have its metadata recorded for reverse lookups. Teach the history helpers to consult that metadata when finding original commits or unapplied filters, returning clear errors when the resolver or its records are missing.

Extend the parser and pretty-printer to understand `:computed` (with optional arguments), and cover the resolver lifecycle, caching, and failure modes with new tests that rely on a tempfile-backed repository.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1932,6 +1932,7 @@ dependencies = [
  "serde_yaml",
  "sled",
  "strfmt",
+ "tempfile",
  "tracing",
 ]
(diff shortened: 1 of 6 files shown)


christian-schilling:
Implemented as #1555

PR #1516, opened 2025-10-07
Upgrade hyper from 0.14.x to 1.x.x

LMG:
This new attempt reimplements hyper-reverse-proxy as a local crate to preserve LFS functionnality (as well as the UI testing feature)


josh-proxy/src/juniper_hyper.rs
```diff
@@ -392,60 +392,99 @@ mod tests {
...
+        let listener = TcpListener::bind(addr).await.unwrap();
+        println!("Listening on http://{}", addr);
+        tokio::task::spawn(async move {
+            loop {
+                let mut rx = shutdown_rx.resubscribe();
+                let (tcp, remote_address) = tokio::select! {
+                    res = listener.accept() => {
+                        match res {
+                            Ok((a,b)) => (a,b),
+                            Err(e) =>{ println!("Error accepting connection: {:?}", e);
```
vlad-ivanov-name, line 440: formatting is messed up here

LMG: I wonder why cargo fmt didn´t catch it

PR #1529, opened 2025-10-14
Fix chain(prefix(a/b),subdir(a))

christian-schilling:
The implementation here was wrong in that it assumed subdir would be deeper than prefix.
Strangely the bug never manifested except now in a scenario of building filters programatically.
I could not get the bug to reproduce by writing a filter spec, so the regression test needs to be
a rust test.

Change: fix-chain-bug


josh-core/src/filter/opt.rs
```diff
@@ -481,6 +485,7 @@ fn step(filter: Filter) -> Filter {
         _ => to_op(filter),
     });
 
+    /* println!("step: {:?} -> {:?}", original, result); */
```
vlad-ivanov-name: leftover stuff

PR #1541, opened 2025-10-24
Fix UI

LMG:
The recent upgrade to hyper 1.0 seems to have broken the UI. This commit fixes the issue with hyper_staticfile. It also changes the return types from a fixed Response<Full<Bytes>> to a custom JoshResponse type that allows us to deal with streaming responses instead of always buffering them.
This type and some helper functions are defined in a new hyper_integration.rs file

### DIFF
--- a/tests/proxy/ui.t
+++ b/tests/proxy/ui.t
@@ -1,0 +1,30 @@
+  $ . ${TESTDIR}/setup_test_env.sh
+  $ cd ${TESTTMP}
+  $ curl -s -I http://127.0.0.1:8002/
+  HTTP/1.1 302 Found\r (esc)
+  location: /~/ui/\r (esc)
+  date: * (glob)
+  \r (esc)
+  $ curl -s -I http://127.0.0.1:8002/~/ui/index.html
+  HTTP/1.1 200 OK\r (esc)
+  etag: * (glob)
+  last-modified: * (glob)
+  accept-ranges: bytes\r (esc)
+  content-length: 633\r (esc)
+  content-type: text/html\r (esc)
+  date: * (glob)
+  \r (esc)
+  $ curl -s -I http://127.0.0.1:8002/~/ui/favicon.ico
+  HTTP/1.1 200 OK\r (esc)
+  etag: * (glob)
+  last-modified: * (glob)
+  accept-ranges: bytes\r (esc)
+  content-length: 12014\r (esc)
+  content-type: image/x-icon\r (esc)
+  date: * (glob)
+  \r (esc)
+  $ curl -s -I http://127.0.0.1:8002/a/repo
+  HTTP/1.1 302 Found\r (esc)
+  location: /~/ui/browse?repo=/a/repo.git&path=&filter=%3A%2F&rev=HEAD\r (esc)
+  date: * (glob)
+  \r (esc)


tests/proxy/ui.t
```diff
@@ -0,0 +1,50 @@
+  $ . ${TESTDIR}/setup_test_env.sh
+  $ cd ${TESTTMP}
+  $ wget http://127.0.0.1:8002/ && cat index.html && rm index.html
+  Connecting to 127.0.0.1:8002 (127.0.0.1:8002)
+  saving to 'index.html'
+  index.html           100% |********************************|   633  0:00:00 ETA
+  'index.html' saved
+  <!doctype html><html lang="en"><head><meta charset="utf-8"/><meta name="viewport" content="width=device-width,initial-scale=1"/><meta name="theme-color" content="#000000"/><meta name="description" content="Web site created using create-react-app"/><link rel="manifest" href="/~/ui/manifest.json"/><link rel="icon" type="image/x-icon" href="/~/ui/favicon.ico"><title>React App</title><script defer="defer" src="/~/ui/static/js/main.36e23a1c.js"></script><link href="/~/ui/static/css/main.4a26a351.css" rel="stylesheet"></head><body><noscript>You need to enable JavaScript to run this app.</noscript><div id="root"></div></body></html> (no-eol)
```
vlad-ivanov-name: tests fail because some resource names (compiled JS) changes. it would perhaps be better to show headers instead of showing full body, and show first e.g. 100 chars/bytes from body with cut or something

tests/proxy/ui.t
```diff
@@ -0,0 +1,50 @@
...
+  $ wget http://127.0.0.1:8002/~/ui/favicon.ico && cat favicon.ico && rm favicon.ico
+  Connecting to 127.0.0.1:8002 (127.0.0.1:8002)
+  saving to 'favicon.ico'
+  favicon.ico          100% |********************************| 12014  0:00:00 ETA
+  'favicon.ico' saved
+  \x00\x00\x01\x00\x03\x0000\x00\x00\x01\x00\x18\x00\xa8\x1c\x00\x006\x00\x00\x00  \x00\x00\x01\x00\x18\x00\xa8\x0c\x00\x00\xde\x1c\x00\x00\x10\x10\x00\x00\x01\x00\x08\x00h\x05\x00\x00\x86)\x00\x00(\x00\x00\x000\x00\x00\x00`\x00\x00\x00\x01\x00\x18\x00\x00\x00\x00\x00\x00\x1b\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x05\x04\x02bV+\x7fq9oe4d^2b]1b]1d^1jc2of2sh2xk5\x82t<\x8d|@\x92\x80A\x95\x84Dzl7\r (no-eol) (esc)
```
vlad-ivanov-name: especially here - cut to a few bytes and use hexdump -C

tests/proxy/ui.t
```diff
@@ -0,0 +1,50 @@
...
+  8Q\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'^),#\x16\x00\x00\x00\x00\x00\x00\x00\x00\x00\x001mX6?$\x05\x00\x00\x00\x00\x00\x00\x00\x00\x00].jGT&>\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00e;=\x17\\F\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x12I\x1bJH*%\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00 (no-eol) (esc)
+  $ wget http://127.0.0.1:8002/a/repo && cat repo
+  Connecting to 127.0.0.1:8002 (127.0.0.1:8002)
+  saving to 'repo'
+  repo                 100% |********************************|   633  0:00:00 ETA
```
vlad-ivanov-name: wget progress bar is probably not very useful in those tests; there could be an option to disable it, or just use curl

vlad-ivanov-name (approved):
\r newlines in cram tests look weird 🤔 other than that lgtm

we should probably swap for e.g. axum at some point to not have to deal with raw hyper, but it's good to have this fixed until then

LMG:
I agree about switching to something higher-level, it's something I'd like to investigate in the future.

PR #1549, opened 2025-10-30
Fix panic on trying to get ~/ui path

LMG:


josh-proxy/src/bin/josh-proxy.rs
```diff
@@ -1209,6 +1209,13 @@ async fn call_service(
     };
 
     if let Some(resource_path) = path.strip_prefix("/~/ui") {
+        if resource_path == "" {
+            let redirect_path = "/~/ui/".to_string();
+            return Ok(Response::builder()
+                .status(hyper::StatusCode::FOUND)
+                .header("Location", redirect_path)
```
vlad-ivanov-name: consider using hyper::header::LOCATION constant

PR #1550, opened 2025-10-30
Fix tracing

### LINKED ISSUE #1548: Broken semantic tracing for HTTP requests
Previously, HTTP requests were traced according to opentelemetry semantic conventions:

Image

This was removed in hyper upgrade PR.

The replacement tracing setup for HTTP requests doesn't seem to work for me, at least on master -- I don't see any traces from HTTP stack at all (e.g. call_service) with default trace levels.

LMG:
Fixes #1548

### DIFF
--- a/josh-proxy/src/bin/josh-proxy.rs
+++ b/josh-proxy/src/bin/josh-proxy.rs
@@ -1548,12 +1548,14 @@ async fn prepare_namespace(
 }
 
 fn trace_http_response_code(trace_span: Span, http_status: StatusCode) {
+    use opentelemetry_semantic_conventions::trace::HTTP_RESPONSE_STATUS_CODE;
+
     macro_rules! trace {
         ($level:expr) => {{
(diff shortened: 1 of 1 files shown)


josh-proxy/src/bin/josh-proxy.rs
```diff
@@ -1581,6 +1590,41 @@ fn make_upstream(remotes: &Vec<cli::Remote>) -> josh::JoshResult<JoshProxyUpstre
...
+        match call_service(proxy_service, req_auth)
+            .instrument(span.clone())
+            .await
+        {
+            Ok(r) => r,
+            Err(e) => make_response(
+                match e {
+                    JoshError(s) => s,
+                }
+                .as_str(),
```
vlad-ivanov-name: the String field in JoshError is public so you can probably just write &e.0

vlad-ivanov-name (approved):
thank you for fixing!

PR #1551, opened 2025-11-01
Refactor apply to take a struct with metadata

christian-schilling:
Change: apply-metadata

### DIFF
--- a/josh-core/src/filter/mod.rs
+++ b/josh-core/src/filter/mod.rs
@@ -1,5 +1,4 @@
 use super::*;
-use history::RewriteData;
 use pest::Parser;
 use std::path::Path;
 mod opt;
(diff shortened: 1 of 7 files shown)


josh-core/src/filter/mod.rs
```diff
@@ -1046,72 +1145,98 @@ fn apply2<'a>(
...
+                .get_path(path)
+                .and_then(|x| repo.find_tree(x.id()))
+                .unwrap_or_else(|_| tree::empty(repo)),
+        )),
+        Op::Prefix(path) => Ok(x.clone().with_tree(tree::insert(
+            repo,
+            &tree::empty(repo),
+            path,
+            x.tree().id(),
+            0o0040000,
```
vlad-ivanov-name, line 1176: i will create an issue for copilot to replace these with constants from git2

PR #1553, opened 2025-11-02
Enforce single commit PRs

christian-schilling:
Change: single-commit-pr


vlad-ivanov-name (approved):
lgtm but note this will require proper PR stacking -- multi-commit PRs that josh split push creates will fail CI

PR #1561, opened 2025-11-03
Implement "pin" filter

vlad-ivanov-name:


christian-schilling:
Dropping in favor of #1573

PR #1584, opened 2025-11-11
Skip changes without id in stacked push

christian-schilling:
Instead of rejecting the whole push


vlad-ivanov-name:
Note: if we have access to sideband channel in future, maybe we can print something there about skipped changes

PR #1588, opened 2025-11-13
Add explicit options for orphan handling in push

christian-schilling:
If a push introduces new orphan history, there is some ambiguity what that means. It could be a synthetic workspace merge commit, an attempt to import an entirely new subtree or just a new history that has been added to an existing subtree.
Previously the "workspace edit" case was assumed and "-p merge" could be used in the "add new subtree" case, leaving no option for the "new history in existing subtree" case.
Instead of guessing this change makes the push fail by default and asks the user to be explicit in what to do.

Also change `merge_base_many` to `merge_base_octopus` as this is the proper function to check for a common ancestor of multiple commits according to git2 docs.

### DIFF
--- a/josh-core/Cargo.toml
+++ b/josh-core/Cargo.toml
@@ -35,3 +35,4 @@ serde_json = { workspace = true }
 serde_yaml = { workspace = true }
 sled = "0.34.7"
 tracing = { workspace = true }
+unindent = "0.2.3"
--- a/tests/proxy/workspace_edit_commit.t
+++ b/tests/proxy/workspace_edit_commit.t
@@ -159,6 +159,21 @@
 
   $ git push origin HEAD:refs/for/master 2>&1 >/dev/null | sed -e 's/[ ]*$//g'
   remote: josh-proxy: pre-receive hook
+  remote: upstream: response status: 500 Internal Server Error
+  remote: upstream: response body:
+  remote:
+  remote: Rejecting new orphan branch at "Add new folders" (5645805dcc75cfe4922b9cb301c40a4a4b35a59d)
+  remote: Specify one of these options:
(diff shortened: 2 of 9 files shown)


RalfJung:
Thanks, this looks great!

So `-o allow_orphans` is what we'd use in the Rust use case where we just want it to sync the history as-is?

christian-schilling:
> Thanks, this looks great!
> 
> So `-o allow_orphans` is what we'd use in the Rust use case where we just want it to sync the history as-is?

Yes, exactly. You *could* just pass it always, but I would say it's probably a good thing to put a speedbump there because adding history with a new orphan is rare and can easily be unintended or done in a wrong / suboptimal way.

josh-core/Cargo.toml
```diff
@@ -35,3 +35,4 @@ serde_json = { workspace = true }
 serde_yaml = { workspace = true }
 sled = "0.34.7"
 tracing = { workspace = true }
+unindent = "0.2.3"
```
vlad-ivanov-name, line 38: we already have indoc in dependencies here, it does the same and also includes format support

tests/proxy/workspace_edit_commit.t
```diff
@@ -159,6 +159,21 @@
 
   $ git push origin HEAD:refs/for/master 2>&1 >/dev/null | sed -e 's/[ ]*$//g'
   remote: josh-proxy: pre-receive hook
+  remote: upstream: response status: 500 Internal Server Error
```
vlad-ivanov-name, line 162: created issue for later https://github.com/josh-project/josh/issues/1592

vlad-ivanov-name:
@claude create a new pull request where you remove the unindent dependency, and replace usage of unindent with indoc's formatdoc! macro

PR #1591, opened 2025-11-13
Implement "josh link" feature

christian-schilling:


josh-cli/src/bin/josh.rs
```diff
@@ -803,6 +854,350 @@ fn handle_push(args: &PushArgs) -> anyhow::Result<()> {
...
+    // Check if we're in a git repository
+    let repo = git2::Repository::open_from_env().context("Not in a git repository")?;
+
+    // Validate the path (should not be empty and should be a valid path)
+    if args.path.is_empty() {
+        return Err(anyhow::anyhow!("Path cannot be empty"));
+    }
+
+    // Normalize the path by removing leading and trailing slashes
+    let normalized_path = args.path.trim_matches('/').to_string();
```
LMG, line 879: Maybe we would want to use https://doc.rust-lang.org/std/fs/fn.canonicalize.html here?

josh-cli/src/bin/josh.rs
```diff
@@ -803,6 +854,350 @@ fn handle_push(args: &PushArgs) -> anyhow::Result<()> {
...
+    // Create the .josh-link.toml content
+    let link_content = toml::to_string(&link_file).context("Failed to serialize link file")?;
+
+    // Create the blob for the .josh-link.toml file
+    let link_blob = repo
+        .blob(link_content.as_bytes())
+        .context("Failed to create blob")?;
+
+    // Create the path for the .josh-link.toml file
+    let link_path = std::path::Path::new(&normalized_path).join(".josh-link.toml");
```
LMG, line 938: Why should it be a hidden file? In the end this is quite similar to a workspace.josh file and I don't see why it should be hidden

josh-cli/src/bin/josh.rs
```diff
@@ -803,6 +854,350 @@ fn handle_push(args: &PushArgs) -> anyhow::Result<()> {
...
+            &signature,
+            &signature,
+            &format!("Add link: {}", normalized_path),
+            &new_tree,
+            &[&head_commit],
+        )
+        .context("Failed to create commit")?;
+
+    // Apply the :link filter to the new commit
+    let link_filter = josh_core::filter::parse(":link=snapshot")
```
LMG, line 969: I guess this mode isn't implemented yet? Is it on purpose that we don´t use embedded here?

josh-cli/src/bin/josh.rs
```diff
@@ -803,6 +854,350 @@ fn handle_push(args: &PushArgs) -> anyhow::Result<()> {
...
+        .map_err(from_josh_err)
+        .context("Failed to parse :link filter")?;
+
+    // Load the cache and create transaction
+    let repo_path = repo.path().parent().unwrap();
+
+    josh_core::cache_sled::sled_load(&repo_path).unwrap();
```
LMG, line 976: In general I don't like unwrap(). Maybe it's okay in a CLI tool, but why not handle the error like for the other error cases, adding some context to what happened?

josh-cli/src/bin/josh.rs
```diff
@@ -803,6 +854,350 @@ fn handle_push(args: &PushArgs) -> anyhow::Result<()> {
...
+        let link_file: JoshLinkFile =
+            toml::from_str(link_content).context("Failed to parse .josh-link.toml")?;
+
+        vec![(std::path::PathBuf::from(path), link_file)]
+    } else {
+        // No path specified - find all .josh-link.toml files in the tree
+        josh_core::find_link_files(&repo, &head_tree)
+            .map_err(from_josh_err)
+            .context("Failed to find link files")?
+    };
```
LMG, line 1058: maybe would be nice for the future to update all paths in a particular directory, i.e. josh link fetch third_party/

josh-cli/src/bin/josh.rs
```diff
@@ -803,6 +854,350 @@ fn handle_push(args: &PushArgs) -> anyhow::Result<()> {
...
+                format!(
+                    "Failed to insert link file into tree at path '{}'",
+                    path.display()
+                )
+            })?;
+    }
+
+    // Create a new commit with the updated tree
+    let signature = if let Ok(time) = std::env::var("JOSH_COMMIT_TIME") {
+        git2::Signature::new(
```
LMG, line 1126: Just wondering, we are always signing commits here, how will that interact with the unsign filter?

josh-core/src/filter/mod.rs
```diff
@@ -1154,6 +1496,106 @@ fn apply2<'a>(transaction: &'a cache::Transaction, op: &Op, x: Apply<'a>) -> Jos
         Op::Linear => Ok(x),
         Op::Prune => Ok(x),
         Op::Unsign => Ok(x),
+        #[cfg(feature = "incubating")]
+        Op::Adapt(adapter) => {
+            let mut result_tree = x.tree().clone();
+            match adapter.as_ref() {
+                "submodules" => {
```
LMG, line 1503: This feels like this should be something that can be implemented for anything in a more configurable way, instead of being hard-coded for a small number of things. You could want to implement it for a package manager or other.

PR #1594, opened 2025-11-15
Extend file filter to support renames

christian-schilling:
This required quite a tricky extension of the syntax, so it's also adding quite a lot of new test cases for parsing the file filter variants.

### DIFF
--- a/josh-core/src/filter/mod.rs
+++ b/josh-core/src/filter/mod.rs
@@ -212,6 +212,11 @@ pub fn message(m: &str) -> Filter {
     to_filter(Op::Message(m.to_string(), MESSAGE_MATCH_ALL_REGEX.clone()))
 }
 
+pub fn file(path: impl Into<std::path::PathBuf>) -> Filter {
+    let p = path.into();
+    to_filter(Op::File(p.clone(), p))
+}
+
 pub fn hook(h: &str) -> Filter {
     to_filter(Op::Hook(h.to_string()))
 }
@@ -303,7 +308,7 @@ enum Op {
     Index,
     Invert,
 
-    File(std::path::PathBuf),
+    File(std::path::PathBuf, std::path::PathBuf), // File(dest_path, source_path)
     Prefix(std::path::PathBuf),
     Subdir(std::path::PathBuf),
     Workspace(std::path::PathBuf),
@@ -619,7 +624,17 @@ fn spec2(op: &Op) -> String {
         Op::Linear => ":linear".to_string(),
         Op::Unsign => ":unsign".to_string(),
         Op::Subdir(path) => format!(":/{}", parse::quote_if(&path.to_string_lossy())),
-        Op::File(path) => format!("::{}", parse::quote_if(&path.to_string_lossy())),
+        Op::File(dest_path, source_path) => {
+            if source_path == dest_path {
+                format!("::{}", parse::quote_if(&dest_path.to_string_lossy()))
+            } else {
+                format!(
+                    "::{}={}",
+                    parse::quote_if(&dest_path.to_string_lossy()),
+                    parse::quote_if(&source_path.to_string_lossy())
+                )
+            }
+        }
         Op::Prune => ":prune=trivial-merge".to_string(),
         Op::Prefix(path) => format!(":prefix={}", parse::quote_if(&path.to_string_lossy())),
         Op::Pattern(pattern) => format!("::{}", parse::quote_if(pattern)),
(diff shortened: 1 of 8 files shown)


josh-core/src/filter/persist.rs
```diff
@@ -423,6 +434,26 @@ fn from_tree2(repo: &git2::Repository, tree_oid: git2::Oid) -> JoshResult<Op> {
...
+                inner
+                    .get_name("1")
+                    .ok_or_else(|| josh_error("message: missing regex"))?
+                    .id(),
+            )?;
+            let fmt = std::str::from_utf8(fmt_blob.content())?.to_string();
+            let regex_str = std::str::from_utf8(regex_blob.content())?;
+            let regex = regex::Regex::new(regex_str)
+                .map_err(|e| josh_error(&format!("invalid regex: {}", e)))?;
+            Ok(Op::Message(fmt, regex))
```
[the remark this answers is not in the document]
christian-schilling: So far the deserialisation code has never been actually used and the filter identity is still derived from the textual representation. So we don't need backward compatibility.

josh-core/src/filter/mod.rs
```diff
@@ -304,7 +308,7 @@ enum Op {
     Index,
     Invert,
 
-    File(std::path::PathBuf),
+    File(std::path::PathBuf, std::path::PathBuf), // File(dest_path, source_path)
```
vlad-ivanov-name, line 311: nit: might want to use {} syntax with named fields

josh-core/src/filter/mod.rs
```diff
@@ -213,6 +213,10 @@ pub fn message(m: &str) -> Filter {
     to_filter(Op::Message(m.to_string(), MESSAGE_MATCH_ALL_REGEX.clone()))
 }
 
+pub fn file(path: std::path::PathBuf) -> Filter {
```
vlad-ivanov-name: since you're creating this "convenience wrapper", you can accept a more flexible input here. for example,

```
pub fn file(path: impl AsRef<str>) -> Filter {
```

or at least &std::path::Path -- I don't think ownership is needed here

josh-core/src/filter/parse.rs
```diff
@@ -87,16 +87,39 @@ fn parse_item(pair: pest::iterators::Pair<Rule>) -> JoshResult<Op> {
         Rule::filter_presub => {
             let mut inner = pair.into_inner();
             let arg = &unquote(inner.next().unwrap().as_str());
+            let has_second_arg = inner.peek().is_some();
+            let second_arg = inner.next().map(|x| unquote(x.as_str()));
```
vlad-ivanov-name: why peek? wouldn't it be:

```
let has_second_arg = second_arg.is_some();
```

josh-core/src/filter/parse.rs
```diff
@@ -87,16 +87,39 @@ fn parse_item(pair: pest::iterators::Pair<Rule>) -> JoshResult<Op> {
...
+                        "Pattern filters cannot use destination=source syntax: {}",
+                        arg
+                    )));
+                }
                 Ok(Op::Pattern(arg.to_string()))
             } else {
-                Ok(Op::File(Path::new(arg).to_owned()))
+                // File case - error if source contains * (patterns not supported in source)
+                if let Some(ref source_arg) = second_arg {
+                    if source_arg.contains('*') {
```
vlad-ivanov-name: you can combine those conditions https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=a46d3092b60086abe16360a89aa3bd7d

josh-core/src/filter/persist.rs
```diff
@@ -454,9 +465,37 @@ fn from_tree2(repo: &git2::Repository, tree_oid: git2::Oid) -> JoshResult<Op> {
             Ok(Op::Prefix(std::path::PathBuf::from(path)))
         }
         "file" => {
-            let blob = repo.find_blob(entry.id())?;
-            let path = std::str::from_utf8(blob.content())?;
-            Ok(Op::File(std::path::PathBuf::from(path)))
+            // Try to read as tree (new format with destination path)
+            if let Ok(inner) = repo.find_tree(entry.id()) {
```
vlad-ivanov-name, line 469: this subtly consumes a bunch of different errors, i guess a more "defensive" approach would be to find object and get its type first. up to you though

josh-core/src/filter/mod.rs
christian-schilling: We do need ownership here, as the Op::File takes PathBuf. If we where to use &Path here that would mean if we already have a PathBuf when calling we need to convert twice PathBuf -> &Path -> PathBuf. As opposed to moving the value. Probably makes no difference in the end 🤔

josh-core/src/filter/mod.rs
christian-schilling: I think that would make it harder to generalize into the "filter categories" like we discussed..

josh-core/src/filter/persist.rs
christian-schilling: I think it's ok for now. Once we have "filter categories" this code should skrink quite a bit and we can also look into more hardening.

PR #1607, opened 2025-11-17
Switch from create-react-app to Vite

LMG:
CRA is deprecated and unmaintained for months
Update tests to vitest so that npm test works (even though we don't use it)
Switch eslint setup to support vite


tests/cli/clone-main-branch.t
```diff
@@ -72,11 +72,11 @@ Test josh clone with main branch as default
 
   $ cd filtered-repo
   $ find .git | grep HEAD
```
vlad-ivanov-name, line 74: Looks like we need to pipe the lines through "sort"

LMG: I dropped it for now, this is not relevant to this change and cli tests don't run in CI

PR #1611, opened 2025-11-17
Implement stored filter nesting

christian-schilling:


josh-core/src/filter/mod.rs
```diff
@@ -23,6 +23,10 @@ static ANCESTORS: LazyLock<
     std::sync::Mutex<std::collections::HashMap<git2::Oid, std::collections::HashSet<git2::Oid>>>,
 > = LazyLock::new(|| std::sync::Mutex::new(std::collections::HashMap::new()));
```
vlad-ivanov-name: Nit: you can use Default::default inside the closure passed to LazyLock

PR #1618, opened 2025-11-22
Add "incubating" feature flag

christian-schilling:
This feature flag is intended to support merging code that is not considered ready to commit to backward compatibility.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -53,3 +53,25 @@ jobs:
           ( cd josh-ssh-dev-server ; go build -o "${CARGO_TARGET_DIR}/josh-ssh-dev-server" )
           sh run-tests.sh --verbose tests/filter/**.t
           sh run-tests.sh --verbose tests/proxy/**.t
+          sh run-tests.sh --verbose tests/cli/**.t
+
+    - name: Run tests (incubating)
+      uses: addnab/docker-run-action@v3
+      with:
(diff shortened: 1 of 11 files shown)


.github/workflows/rust.yml
```diff
@@ -53,3 +53,25 @@ jobs:
...
+    - name: Run tests (incubating)
+      uses: addnab/docker-run-action@v3
+      with:
+        image: josh-ci-dev:latest
+        options: -v ${{ github.workspace }}:/github/workspace -w /github/workspace
+        run: |
+          set -e -x
+
+          # Formatting
+          cargo fmt -- --check
```
LMG, line 67: this could be done just once

.github/workflows/rust.yml
```diff
@@ -53,3 +53,25 @@ jobs:
           ( cd josh-ssh-dev-server ; go build -o "${CARGO_TARGET_DIR}/josh-ssh-dev-server" )
           sh run-tests.sh --verbose tests/filter/**.t
           sh run-tests.sh --verbose tests/proxy/**.t
+          sh run-tests.sh --verbose tests/cli/**.t
+
+    - name: Run tests (incubating)
```
vlad-ivanov-name, line 58: Should this be a matrix? 

https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations

PR #1647, opened 2025-12-04
Update gix crates and run cargo update

LMG:


vlad-ivanov-name:
Updated in a different PR now

PR #1648, opened 2025-12-07
Fix message filter being put in wrong position by optimizer

vlad-ivanov-name:


josh-core/src/filter/opt.rs
```diff
@@ -315,6 +315,8 @@ fn common_post(filters: &Vec<Filter>) -> Option<(Filter, Vec<Filter>)> {
             common_post.map(|c| (c, rest))
         } else if let Op::Prefix(_) = to_op(c) {
             common_post.map(|c| (c, rest))
+        } else if let Op::Message(..) = to_op(c) {
```
christian-schilling, line 318: I wonder why this was limited to prefix in the first place. Maybe we should look into allowing factorizing all filters here instead of adding more special cases 🤔 
Thoughts?

vlad-ivanov-name: Would prefer to merge this as a hotfix and then create an issue. For all filters I think we'd need a good test strategy, maybe something like guided fuzzing to discover "unsound" edge cases

christian-schilling: Makes sense

christian-schilling (approved):

PR #1665, opened 2025-12-13
Ensure find_unapply_base exits early

vlad-ivanov-name:


josh-core/src/history.rs
```diff
@@ -110,18 +111,112 @@ fn find_unapply_base(
     walk.set_sorting(git2::Sort::TOPOLOGICAL)?;
     walk.push(contained_in)?;
 
+    // libgit2 does not read the list of commits lazily; instead,
+    // it reads the whole list, sorts it according to user's preference,
+    // and then lets the user iterate.
+    //
+    // while this does ensure the desired iteration order, it also means
+    // revwalks don't really scale. this approach below abuses "hide callbacks"
```
christian-schilling: Why "abuse"? Seem just like legit use to me...

vlad-ivanov-name: will remove, but i initially felt it's a misuse of it because it's happening before sort, and there's a dozen different ways this could affect the commit lists depending on the order it's being called. but i guess this is defined somewhere so whatever

josh-core/src/history.rs
```diff
@@ -110,18 +111,112 @@ fn find_unapply_base(
...
+        // we don't add more commits to walk list as the loop
+        // in libgit2 continues if callback returns true
+        //
+        // https://github.com/libgit2/libgit2/blob/610dcaac065c0f27daca84b87795ed26926864f5/src/libgit2/revwalk.c#L428-L429
+        if ctx.result.is_some() {
+            return true;
+        }
+
+        ctx.pending.push(oid);
+        if !ctx.unlocked.contains(&oid) {
```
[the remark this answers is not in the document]
vlad-ivanov-name: this is incorrect -- the loop below will remove the commit from pending list immediately if it's processable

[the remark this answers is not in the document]
vlad-ivanov-name: replaced with hashset

josh-core/src/history.rs
```diff
@@ -110,18 +111,112 @@ fn find_unapply_base(
...
+                .position(|&pending_oid| ctx.unlocked.contains(&pending_oid))
+            {
+                Some(idx) => idx,
+                None => break,
+            };
+
+            let pending_oid = ctx.pending.swap_remove(idx);
+
+            let commit = match transaction.repo().find_commit(pending_oid) {
+                Ok(oid) => oid,
```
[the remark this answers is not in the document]
vlad-ivanov-name: done

josh-core/src/history.rs
```diff
@@ -110,18 +111,112 @@ fn find_unapply_base(
...
+
+    let walk = walk.with_hide_callback(&mut hide_callback)?;
     for original in walk {
-        let original = transaction.repo().find_commit(original?)?;
-        if filtered == filter::apply_to_commit(filter, &original, transaction)? {
-            // In case a match is found, cache the result
-            filtered_to_original.insert(filtered, original.id());
-            tracing::info!("found original properly {}", original.id());
-            return Ok(original.id());
+        let _original = original?;
```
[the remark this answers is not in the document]
vlad-ivanov-name: done

josh-core/src/history.rs
vlad-ivanov-name: done

josh-core/src/history.rs
```diff
@@ -110,18 +111,114 @@ fn find_unapply_base(
...
+
+    let walk = walk.with_hide_callback(&mut hide_callback)?;
     for original in walk {
-        let original = transaction.repo().find_commit(original?)?;
-        if filtered == filter::apply_to_commit(filter, &original, transaction)? {
-            // In case a match is found, cache the result
-            filtered_to_original.insert(filtered, original.id());
-            tracing::info!("found original properly {}", original.id());
-            return Ok(original.id());
+        // Only propagate errors; value is unused. Drives iteration to trigger hide_callback.
```
christian-schilling, line 202: I think you can just to walk.count() to trigger iteration.

vlad-ivanov-name: true but i'd like to propagate errors still -- like this we rely on implementation details of libgit2 a bit less.

christian-schilling (approved):

PR #1689, opened 2025-12-21
Avoid double-hashing filters

vlad-ivanov-name:
This brings a minor but measurable perf improvement 

image

### DIFF
--- a/josh-core/src/filter/hash.rs
+++ b/josh-core/src/filter/hash.rs
@@ -1,0 +1,82 @@
+use std::hash::Hasher;
+
+/// A hasher that uses the first 8 bytes of a SHA1 hash directly as the hash value.
+/// This is designed to be used with HashMaps where the key is already a SHA1 hash,
+/// e.g. Filter or Oid, avoiding double-hashing.
+#[derive(Default)]
+pub struct PassthroughHasher {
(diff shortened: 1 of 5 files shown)


josh-core/src/filter/hash.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    // hyper-specialized: reject everything that's not sha1 length
+    fn write(&mut self, bytes: &[u8]) {
+        if self.done {
+            panic!("hasher data already written");
+        }
+
+        // skip length prefix
+        if bytes.len() == size_of::<usize>() {
```
vlad-ivanov-name, line 29: this could be done via overriding write_length_prefix, but that feature is nightly only.

christian-schilling: Maybe add a FIXME for when the rust feature lands in stable?

christian-schilling:
We have some more HashMaps...change them to? Or add benches first?

vlad-ivanov-name:
those are mostly less "hot" so benchmark first would make sense

josh-core/src/filter/hash.rs
vlad-ivanov-name: done

PR #1704, opened 2025-12-23
feat: use reqwest for josh-proxy auth

Minion3665:
There are various benefits to using reqwest rather than directly using hyper comprising:

- A simpler and, in my opinion, easier-to-read interface
- We already use reqwest elsewhere in the code (in juniper_hyper tests and josh-ssh-shell)
- Reqwest handles http responses better. For example, it follows redirects which are used by forges such at tangled extensively (<https://tangled.org/>). Tangled in particular has repositories ending in `.git` redirected to URLs that do not end in `.git`, which breaks Josh

(N.B. It was previously possible to workaround this with a git config providing `insteadOf` URLs like so:

    [url "https://tangled.org/freshlybakedca.ke/patisserie"]
      insteadOf = "https://tangled.org/freshlybakedca.ke/patisserie.git"

, but an update to either Josh or
Tangled broke this again. I'm not quite sure which...)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3170,16 +3170,20 @@ dependencies = [
  "http-body",
  "http-body-util",
  "hyper",
+ "hyper-tls",
  "hyper-util",
  "js-sys",
  "log",
+ "native-tls",
  "percent-encoding",
  "pin-project-lite",
+ "rustls-pki-types",
  "serde",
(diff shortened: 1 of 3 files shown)


vlad-ivanov-name:
Thank you for the contribution and the detailed description!

PR #1705, opened 2025-12-23
Add compose_many; use filter builder in benchmark

vlad-ivanov-name:


josh-core/src/build.rs
```diff
@@ -7,3 +7,7 @@ use crate::filter::persist::to_filter;
 pub fn compose(first: Filter, second: Filter) -> Filter {
     opt::optimize(to_filter(Op::Compose(vec![first, second])))
 }
+
+pub fn compose_many(filters: &[Filter]) -> Filter {
```
christian-schilling: I think we can call this one `compose` and remove the existing two parameter one.

vlad-ivanov-name: Done

christian-schilling (approved):

PR #1710, opened 2025-12-26
Add filter metadata support

christian-schilling:
We will need to be able to add metadata to filters to customize behaviour (parent handling).

This also changes "josh link" to use the new flags feature instead of toml for the link files.


josh-core/src/filter/mod.rs
```diff
@@ -1174,16 +1239,19 @@ pub fn apply_to_commit2(
 
             /* dbg!("embed"); */
             /* dbg!(&path); */
-            if let Some(link) = read_josh_link(repo, &commit.tree()?, &path, ".josh-link.toml") {
+            if let Some(link) = read_josh_link(repo, &commit.tree()?, &path, ".link.josh") {
```
vlad-ivanov-name, line 1242: you can have less nesting here by using and_then().

josh-core/src/filter/mod.rs
```diff
@@ -2097,4 +2186,90 @@ mod tests {
         // The spec should contain the chain representation
         assert!(!spec_str.is_empty());
     }
+
+    #[test]
+    fn meta_filter_parsing_test() {
```
vlad-ivanov-name, line 2191: these tests look rather messy with all the string escaping going on; consider raw strings.

josh-core/src/filter/persist.rs
```diff
@@ -908,6 +920,40 @@ fn from_tree2(repo: &git2::Repository, tree_oid: git2::Oid) -> JoshResult<Op> {
...
+            let filter_tree = repo.find_tree(
+                meta_tree
+                    .get_name("0")
+                    .ok_or_else(|| josh_error("meta: missing filter tree"))?
+                    .id(),
+            )?;
+
+            // Deserialize metadata map - keys are filenames, values are blob contents
+            let mut meta = std::collections::BTreeMap::new();
+            for i in 0..meta_tree.len() {
```
vlad-ivanov-name, line 934: you could iterate with iter()

Image

if it's about order being significant, comment would help

PR #1729, opened 2025-12-31
Make is_ancestor_of use the sequence number for early out

christian-schilling:
This changes the sequence number to use the max sequence number of the parents instead of the one from the first parent, to guarantee that children always have larger numbers than all of their parents.


josh-core/src/cache/transaction.rs
```diff
@@ -478,37 +478,48 @@ pub fn compute_sequence_number(
...
+    if no_walk {
         transaction.insert(
             crate::filter::sequence_number(),
             commit.id(),
-            oid_from_u128(pc + 1),
+            oid_from_u128(this_sequence_number),
             true,
         );
+    } else {
+        log::info!("compute_sequence_number: new_walk for {:?}", input);
```
vlad-ivanov-name, line 509: should this be log::debug?

josh-core/src/cache/transaction.rs
```diff
@@ -478,37 +478,48 @@ pub fn compute_sequence_number(
...
+        walk.set_sorting(git2::Sort::REVERSE | git2::Sort::TOPOLOGICAL)?;
+        walk.push(input)?;
+
+        for c in walk {
+            let commit = transaction.repo().find_commit(c?)?;
+            let mut this_sequence_number = 0;
+            for parent in commit.parent_ids() {
+                let parent_sequence_number = compute_sequence_number(transaction, parent)?;
+                this_sequence_number =
+                    std::cmp::max(this_sequence_number, parent_sequence_number + 1);
```
vlad-ivanov-name, line 520: it would be great to add PR description as a comment here

josh-core/src/cache/transaction.rs
christian-schilling: I think info is fine. not happening that often

josh-core/src/cache/transaction.rs
```diff
@@ -466,6 +466,9 @@ pub fn u128_from_oid(oid: git2::Oid) -> u128 {
     u128::from_be_bytes(n)
 }
 
+/// Computes the sequence number for each commit so that the sequence_number for any
+/// commit is always larger than the sequence_number of all of it's parents.
```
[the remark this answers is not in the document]
christian-schilling: fixed

PR #1735, opened 2026-01-04
Serve /info/refs by josh-proxy itself

vlad-ivanov-name:
LLM generated PoC code; to be rewritten by hand

---

Instead of spawning http-backend, serve /info/refs by josh. This provides an advantage: previously, we needed to create a namespace where we populate state of refs, in order for git http-backend to pick it up. Essentially, we are using git repo directory as transport for the list of refs. This doesn't scale and is really slow when there are a lot of branches.

This reduces push time; with llvm repo (2.5k refs / 600 branches), best case scenario (warm i/o cache) goes from 5.5s to 3.7s or so. This roughly matches profiling results: creating all the refs alone takes a second or so.


vlad-ivanov-name:
Closed in favor of https://github.com/josh-project/josh/pull/1767

PR #1755, opened 2026-01-09
Implement history linearization as meta option

christian-schilling:


josh-core/src/history.rs
```diff
@@ -19,7 +19,7 @@ pub fn walk2(
 
     let walk = {
         let mut walk = transaction.repo().revwalk()?;
-        if filter::is_linear(filter) {
+        if filter.get_meta("graph") == Some("linear".to_string()) {
```
vlad-ivanov-name: this allocates a string every time.

PR #1764, opened 2026-01-13
Extend :linear syntax to take filter as argument

christian-schilling:


docs/src/reference/filters.md
```diff
@@ -153,7 +153,7 @@ All files matching the pattern in arbitrarily deep subdirectories of the input
 
 These filter do not modify git trees, but instead only operate on the commit graph.
 
-### Linearise history **:linear**
+### Linearise history **:linear[:filter]**
```
vlad-ivanov-name, line 156: not obvious what filter here does

vlad-ivanov-name (approved):

PR #1767, opened 2026-01-14
Handle /info/refs inside proxy itself

vlad-ivanov-name:
This is second version of the previous pull request -- now done properly with most of the refactorings needed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2169,6 +2169,8 @@ dependencies = [
  "futures",
  "git2",
  "gix",
+ "gix-packetline",
+ "gix-transport",
  "hex",
  "indoc",
  "josh-core",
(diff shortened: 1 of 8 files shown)


vlad-ivanov-name:
@claude review this pull request.

(PRs in this repo always should contain just one commit, so when obtaining the diff, look just at the top commit of the branch)

vlad-ivanov-name:
>1. Potential data race in NamespacedRefs::into_inner - josh-proxy/src/service.rs:468

that's fine, intended to do so, we don't need to persist refs

>2. Missing error handling in capability discovery - josh-proxy/src/serve.rs:39-119

since we aren't trying to test git itself here, and the format is rather stable, i don't think this is an issue

>3. Capability discovery uses wrong path - josh-proxy/src/bin/josh-proxy.rs:156

mirror repo is created using create_repo, this is just wrong

>4. Inconsistent ref sorting logic - josh-proxy/src/serve.rs:162-166

the logic is implemented to match git, and this is reflected in tests that didn't change. open to hear ways to ensure this is correct though.

>5. Capabilities are static for the proxy's lifetime - josh-proxy/src/service.rs:130

this is intentional

christian-schilling (approved):

vlad-ivanov-name:
@claude rebase this PR on latest master. Keep this PR to 1 commit.

PR #1774, opened 2026-01-18
Remove trivial merges by default

christian-schilling:
This removes trivial merges that are created as a result of the filtering, but keeps those that have been trivial without the filter. To enable backward compatibility we introduce a "meta option" to get the old behaviour of keeping all trivial merges.
Also the :prune=trivial-merge filter is kept as it can prune trivial merges that are already present in the original history.


josh-core/src/history.rs
```diff
@@ -19,7 +19,7 @@ pub fn walk2(
 
     let walk = {
         let mut walk = transaction.repo().revwalk()?;
-        if filter.get_meta("graph").is_some_and(|s| s == "linear") {
+        if filter.get_meta("history").is_some_and(|s| s == "linear") {
```
vlad-ivanov-name, line 22: maybe add some constants for string literals?

PR #1779, opened 2026-01-21
josh-mq: basic config management

vlad-ivanov-name:


vlad-ivanov-name:
closing -- needs to integrate josh-link, and some refactorings need to happen before that

PR #1782, opened 2026-01-24
Further cleanup of josh-cli + split a crate out of binary

vlad-ivanov-name:


josh-link/src/lib.rs
```diff
@@ -0,0 +1,183 @@
+use anyhow::Context;
+use josh_core::filter::tree;
+use std::path::PathBuf;
+
+/// Result from adding a link to a repository
+pub struct AddLinkResult {
+    /// Commit with .link.josh file added
+    pub commit_with_link: git2::Oid,
+    /// Commit after applying :link=snapshot filter
```
christian-schilling, line 9: We will propably need to change that. We don't always want to use "snapshot".
It should be somehow configurable what mode to use, but I'm not sure how yet.

PR #1784, opened 2026-01-26
Implement prototype starlark evaluation

christian-schilling:


josh-starlark/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "josh-starlark"
+version = "22.4.15"
```
vlad-ivanov-name, line 3: do we need this here? will it go to crates.io etc?

josh-starlark/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+authors = ["Christian Schilling <initcrash@gmail.com>"]
+license-file = "../LICENSE"
+description = "Starlark support for creating josh filters"
+keywords = ["git", "monorepo", "workflow", "scm", "starlark"]
+readme = "../README.md"
+repository = "https://github.com/josh-project/josh"
+
+[dependencies]
+josh-filter = { path = "../josh-filter" }
+starlark = "0.13.0"
```
vlad-ivanov-name, line 14: kind of unfortunate it has no feature flags because it adds like a million dependencies 😬

josh-starlark/src/evaluate.rs
```diff
@@ -0,0 +1,64 @@
...
+};
+use std::sync::{Arc, Mutex};
+
+/// Evaluate a starlark script and return the resulting Filter
+///
+/// The script should define a function or variable that returns a Filter.
+/// The script must not use josh filter language strings - all filters
+/// must be constructed using the Filter builder methods.
+///
+/// The `tree_oid` parameter is made available as a global variable named "tree"
```
vlad-ivanov-name, line 19: not very familiar with starlark but are globals supposed to be all caps?

josh-starlark/src/evaluate.rs
```diff
@@ -0,0 +1,64 @@
...
+/// in the Starlark script, allowing access to the git tree via methods.
+pub fn evaluate(
+    script: &str,
+    tree_oid: git2::Oid,
+    repo: Arc<Mutex<git2::Repository>>,
+) -> anyhow::Result<Filter> {
+    // Parse the starlark script
+    let ast = AstModule::parse("script.star", script.to_owned(), &Dialect::Standard)
```
vlad-ivanov-name, line 27: nit: technically could be tree entry name?

vlad-ivanov-name: or have an optional name arg and unwrap_or to script.star

josh-starlark/src/evaluate.rs
```diff
@@ -0,0 +1,64 @@
...
+    // Evaluate the script
+    let _result = eval
+        .eval_module(ast, &globals)
+        .map_err(|e| anyhow::anyhow!("Failed to evaluate starlark script: {}", e))?;
+
+    // Try to get the filter from the module
+    // Look for a variable named "filter"
+    let filter_value = module.get("filter").ok_or_else(|| {
+        anyhow::anyhow!("Script must define 'filter' variable returning a Filter")
+    })?;
```
vlad-ivanov-name, line 56: again not sure about starlark conventions but would it be better to export a function, if that's possible?

josh-starlark/src/evaluate.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    // Create a new module
+    let module = Module::new();
+
+    // Build globals with our filter module
+    let globals = GlobalsBuilder::standard().with(filter_module).build();
+
+    // Add a global "filter" value (nop filter) to the module
+    let filter_value = module.heap().alloc(StarlarkFilter::new());
+    module.set("filter", filter_value);
```
vlad-ivanov-name, line 38: nit: the comment above doesn't mention filter global

josh-starlark/src/filter.rs
```diff
@@ -0,0 +1,282 @@
...
+    fn is_nop(this: &StarlarkFilter) -> anyhow::Result<bool> {
+        Ok(this.is_nop())
+    }
+    fn peel(this: &StarlarkFilter) -> anyhow::Result<StarlarkFilter> {
+        Ok(this.peel())
+    }
+}
+
+impl StarlarkFilter {
+    /// Create a new Filter
```
vlad-ivanov-name, line 131: the comments here and below are more or less useless as they don't add any new information given the function signature.

but it would be useful to add a comment on why both this is required and filter_methods above with more or less the same functions

josh-starlark/src/tests.rs
```diff
@@ -0,0 +1,280 @@
...
+#[test]
+fn test_simple_filter() {
+    // Create a temporary repository for testing
+    let temp_dir = std::env::temp_dir().join("josh_starlark_test");
+    let _ = std::fs::remove_dir_all(&temp_dir);
+    let repo = git2::Repository::init(&temp_dir).unwrap();
+    let empty_tree_oid = git2::Oid::from_str("4b825dc642cb6eb9a060e54bf8d69288fbee4904").unwrap();
+    let repo_arc = Arc::new(Mutex::new(repo));
+
+    let script = r#"
```
vlad-ivanov-name, line 14: nit: we have indoc in dependencies if that helps.

josh-starlark/src/tests.rs
```diff
@@ -0,0 +1,280 @@
...
+// Helper function to create a test repository with files and directories
+fn create_test_repo() -> (Arc<Mutex<git2::Repository>>, git2::Oid) {
+    let temp_dir = std::env::temp_dir().join(format!(
+        "josh_starlark_tree_test_{}",
+        std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .unwrap()
+            .as_nanos()
+    ));
+    let _ = std::fs::remove_dir_all(&temp_dir);
```
vlad-ivanov-name, line 116: not sure what's happening here; would this dir be cleaned up properly? one option would be to use tempfile::TempDir, it has a drop guard, just need to return it from the function

josh-starlark/src/tests.rs
```diff
@@ -0,0 +1,280 @@
...
+    //   - main.rs (blob)
+    //   - lib/ (tree)
+    //     - utils.rs (blob)
+
+    let lib_tree_oid = {
+        let repo_guard = repo_arc.lock().unwrap();
+        // Create blobs
+        let utils_rs_blob = repo_guard
+            .blob(b"pub fn helper() {\n    // helper\n}")
+            .unwrap();
```
vlad-ivanov-name: consider returning result from this helper

josh-starlark/src/tests.rs
```diff
@@ -0,0 +1,280 @@
...
+fn test_tree_nested_access() {
+    let (repo_arc, root_tree_oid) = create_test_repo();
+
+    // Test nested access: tree.tree("src").file("main.rs")
+    let script = r#"
+src_tree = tree.tree("src")
+main_content = src_tree.file("main.rs")
+filter = filter.subdir("src")
+"#;
+    let filter = evaluate(script, root_tree_oid, Arc::clone(&repo_arc)).unwrap();
```
vlad-ivanov-name: here and in other places: why not just .clone() 🤔

josh-starlark/src/tree.rs
```diff
@@ -0,0 +1,255 @@
...
+            return Ok(self.tree_oid);
+        }
+
+        let repo = self
+            .repo
+            .lock()
+            .map_err(|e| anyhow::anyhow!("Failed to lock repository: {}", e))?;
+
+        let path_buf = PathBuf::from(path);
+        let components: Vec<&str> = path_buf.iter().map(|c| c.to_str().unwrap()).collect();
```
vlad-ivanov-name: if to_str() returns an Option, anyhow has a method .context that adds error message and lets you use `?`

josh-starlark/src/tree.rs
```diff
@@ -0,0 +1,255 @@
...
+
+        let mut current_tree_oid = self.tree_oid;
+        for component in components {
+            let current_tree = repo
+                .find_tree(current_tree_oid)
+                .map_err(|e| anyhow::anyhow!("Failed to find tree: {}", e))?;
```
vlad-ivanov-name: can also be expressed simpler as 

```
.context("Failed to find tree")
```

josh-starlark/Cargo.toml
christian-schilling: I think every crate needs it 🤷🏻‍♂️

josh-starlark/src/evaluate.rs
christian-schilling: Ah, not sure if it matters 🤔

josh-starlark/src/evaluate.rs
christian-schilling: Constants, yes. But those are not constants.
Still not sure to keep it that way though.
Also thinking maybe having a convention to always call a function instead of executing the code at global scope. But don't know, have to learn a bit more about starlark.

josh-starlark/src/evaluate.rs
christian-schilling: Maybe...don't know.. This is all prototype

josh-starlark/src/evaluate.rs
christian-schilling: hm?

josh-starlark/src/tests.rs
christian-schilling: I'll probably move the test to cram anyway..

josh-starlark/src/tests.rs
christian-schilling: I think as soon as tempdir is involved it should be cram anyway. But can do this in followup

josh-starlark/src/evaluate.rs
christian-schilling: Not, sure. Maybe constants, but those are not constants

vlad-ivanov-name:
wrt function vs variable, global names etc -- we can iterate on this, doesn't have to be the final API

PR #1785, opened 2026-01-28
Rework :rev filter

christian-schilling:
This makes the :rev filter support different modes of commit selection. Since it is now based on per_rev_filter it can also cover the use case of the :from filter


josh-core/src/filter/persist.rs
```diff
@@ -132,9 +133,22 @@ impl InMemoryBuilder {
...
+            // Encode match operator as prefix
+            let key = match match_op {
+                RevMatch::AncestorStrict => format!("<{}", lazy_ref),
+                RevMatch::AncestorInclusive => format!("<={}", lazy_ref),
+                RevMatch::Equal => format!("={}", lazy_ref),
+                RevMatch::Default => {
+                    // Default filter uses "_" as key (no SHA)
+                    "_".to_string()
+                }
+            };
```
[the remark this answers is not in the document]
christian-schilling: This is only a problem on load, not on write

josh-core/src/filter/persist.rs
```diff
@@ -749,7 +824,7 @@ fn from_tree2(repo: &git2::Repository, tree_oid: git2::Oid) -> JoshResult<Op> {
         }
         "rev" => {
             let rev_tree = repo.find_tree(entry.id())?;
-            let mut filters = std::collections::BTreeMap::new();
+            let mut filters = Vec::new();
             for i in 0..rev_tree.len() {
                 let entry = rev_tree
                     .get(i)
```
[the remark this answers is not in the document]
christian-schilling: fixed

docs/src/reference/filters.md
```diff
@@ -208,21 +208,42 @@ These filter do not modify git trees, but instead only operate on the commit gra
...
-ancestor also has a matching entry in the `:rev(...)` it's filter will *replace* `<filter_N>`
-for all further ancestors (and so on).
+The `:rev(...)` filter allows you to apply different filters to different parts of the commit history based on commit relationships. Each entry in the filter specifies a condition and a filter to apply when that condition matches.
 
-This special value `0000000000000000000000000000000000000000` can be used as a `<sha_n>` to filter
-commits that don't match any of the other shas.
+**Syntax:**
+```
+:rev(
+  <operator><sha>:filter,
```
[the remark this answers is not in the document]
christian-schilling: fixed

josh-core/src/filter/mod.rs
```diff
@@ -430,8 +430,8 @@ fn lazy_refs2(op: &Op) -> Vec<String> {
             av
         }
         Op::Rev(filters) => {
-            let mut lr = lazy_refs2(&Op::Compose(filters.values().copied().collect()));
-            lr.extend(filters.keys().filter_map(|x| {
+            let mut lr = lazy_refs2(&Op::Compose(filters.iter().map(|(_, _, f)| *f).collect()));
```
vlad-ivanov-name, line 433: not for this PR, but we should probably add a trait like GetLazyRefs and implement it for a bunch of types here -- would move all the ref collection code out of this fn

josh-core/src/filter/mod.rs
```diff
@@ -430,8 +430,8 @@ fn lazy_refs2(op: &Op) -> Vec<String> {
...
+            lr.extend(filters.iter().filter_map(|(_, x, _)| {
```
vlad-ivanov-name, line 434: maybe it's an opportunity to give "x" a more meaningful name

josh-core/src/filter/mod.rs
```diff
@@ -690,41 +670,77 @@ fn read_josh_link<'a>(
...
+                    return Err(josh_error("unresolved lazy ref"));
+                };
 
-    let mut matching_filters = Vec::new();
+                if !transaction.repo().odb()?.exists(*filter_tip) {
+                    return Err(josh_error(&format!(
+                        "`:rev(...)` with nonexistent OID: {}",
+                        filter_tip
+                    )));
+                }
```
vlad-ivanov-name, line 693: this whole block is just repeated three times, you can move it out of match. and you can move call to is_ancestor_of outside too, and just use match arms to check conditions

josh-core/src/filter/opt.rs
```diff
@@ -520,7 +520,12 @@ fn step(filter: Filter) -> Filter {
                 Op::Prefix(path)
             }
         }
-        Op::Rev(filters) => Op::Rev(filters.into_iter().map(|(i, f)| (i, step(f))).collect()),
+        Op::Rev(filters) => Op::Rev(
```
vlad-ivanov-name, line 523: not sure if necessary but one possible optimization is to just drop everything after `_`

josh-core/src/filter/persist.rs
```diff
@@ -768,33 +843,32 @@ fn from_tree2(repo: &git2::Repository, tree_oid: git2::Oid) -> JoshResult<Op> {
                         .id(),
                 )?;
                 let key = std::str::from_utf8(key_blob.content())?.to_string();
+
+                // Parse match operator from key
+                let (match_op, lazy_ref) = if key == "_" {
+                    // Default filter - no SHA needed
+                    (RevMatch::Default, LazyRef::Resolved(git2::Oid::zero()))
+                } else if key.starts_with("<=") {
```
vlad-ivanov-name, line 851: else if let Some(ref_str) = key.strip_prefix("<=")

vlad-ivanov-name: same below

josh-core/src/flang/mod.rs
```diff
@@ -160,11 +160,19 @@ pub(crate) fn spec2(op: &Op) -> String {
...
         }
         Op::Rev(filters) => {
-            let mut v = filters
+            // No sorting - preserve order for first-match semantics
+            let v = filters
                 .iter()
-                .map(|(k, v)| format!("{}{}", k, spec(*v)))
+                .map(|(match_op, k, v)| {
+                    let match_str = match match_op {
+                        RevMatch::AncestorStrict => format!("<{}", k),
```
vlad-ivanov-name, line 168: I think you could implement Display or ToString for RevMatch, then this kind of code would be simplified in several places

josh-core/src/flang/parse.rs
```diff
@@ -181,37 +181,69 @@ fn parse_item(pair: pest::iterators::Pair<Rule>) -> JoshResult<Filter> {
...
-        Rule::filter_from => {
-            let v: Vec<_> = pair.into_inner().map(|x| x.as_str()).collect();
-
-            if v.len() == 2 {
-                let oid = LazyRef::parse(v[0])?;
-                let filter = parse(v[1])?;
-                Ok(filter.chain(filter::to_filter(Op::HistoryConcat(oid, filter))))
-            } else {
-                Err(josh_error("wrong argument count for :from"))
+            let mut entries = Vec::new();
```
vlad-ivanov-name, line 184: maybe time to factor it out because it's already difficult to read and review at this nesting level

josh-core/src/flang/parse.rs
```diff
@@ -181,37 +181,69 @@ fn parse_item(pair: pest::iterators::Pair<Rule>) -> JoshResult<Filter> {
...
+                                    .ok_or_else(|| josh_error("rev_default: missing filter"))?;
+                                let filter = parse(filter_pair.as_str())?;
+                                entries.push((
+                                    RevMatch::Default,
+                                    LazyRef::Resolved(git2::Oid::zero()),
+                                    filter,
+                                ));
+                            }
+                            Rule::rev_match => {
+                                // Regular match with operator, SHA, and filter
```
vlad-ivanov-name, line 207: this should be impl of https://doc.rust-lang.org/std/str/trait.FromStr.html

tests/filter/rev.t
```diff
@@ -51,34 +51,47 @@
   * 9f0db868b59a422c114df33bc6a8b2950f80490b:a087bfbdb1a5bad499b40ccd1363d30db1313f54
 
   $ josh-filter -s ":rev(ffffffffffffffffffffffffffffffffffffffff:prefix=x/y)" --update refs/heads/filtered
-  [5] :prefix=x
-  [5] :prefix=y
-  [10] sequence_number
-  ERROR: `:rev(...)` with nonexistent OID: ffffffffffffffffffffffffffffffffffffffff
+  ERROR: Invalid workspace:
```
vlad-ivanov-name, line 54: do you intend to keep this test this way? would it be better to update to again trigger non existent oid message?

josh-core/src/filter/opt.rs
christian-schilling: I think it's not really necessary...

PR #1786, opened 2026-01-28
Use .josh files to store josh remotes

christian-schilling:


josh-cli/src/bin/josh.rs
```diff
@@ -443,6 +444,135 @@ fn to_absolute_remote_url(url: &str) -> anyhow::Result<String> {
     }
 }
 
+/// Read remote configuration from .git/josh/remotes/<name>.josh file
+/// Falls back to legacy git config josh-remote section if file doesn't exist
```
vlad-ivanov-name, line 448: is anyone even using that? imo we could just delete the old way

tests/cli/remote-refspecs.t
```diff
@@ -27,8 +27,13 @@ Test that josh remote add sets up proper refspecs
   $ git config --get-all remote.origin.fetch
   +refs/heads/*:refs/remotes/origin/*
 
-  $ git config josh-remote.origin.filter
-  :/sub1
+  $ cat .git/josh/remotes/origin.josh | sed "s|file://.*/remote/libs|file://\${TESTTMP}/remote/libs|"
```
vlad-ivanov-name, line 30: might be simpler to use cram/prysk globs

PR #1794, opened 2026-02-01
Store change graph instead of refs

vlad-ivanov-name:


josh-cq/src/bin/josh-cq.rs
```diff
@@ -47,27 +50,24 @@ fn handle_track(
 ) -> anyhow::Result<()> {
     let repo = transaction.repo();
 
-    // Fetch refs from remote
-    let refs = josh_cq::remote::list_refs(&args.url)?;
+    // Fetch everything from the remote
```
christian-schilling, line 57: Why do we need to fetch at this point?
The overal idea is that the remotes content does not get fetched into the metarepo.

josh-cq/src/vendor/generic.rs
```diff
@@ -0,0 +1,158 @@
...
+/// to be considered a change
+pub struct GenericRemote {
+    pub repo_path: std::path::PathBuf,
+    pub url: String,
+    pub trunk_ref: String,
+}
+
+impl GenericRemote {}
+
+fn get_change_id(commit: &git2::Commit) -> Option<ChangeId> {
```
christian-schilling, line 16: We have that code somewhere else already -> dedup?

vlad-ivanov-name: will do later -- will need to either unify code with proxy or a separate crate

josh-cq/src/bin/josh-cq.rs
vlad-ivanov-name: done

PR #1804, opened 2026-02-07
Fix Docker build cache scope for PR branches

vlad-ivanov-name:
LLM generated

---

Configure GHA cache to read from both current branch and main branch, allowing PRs to get cache hits from master builds.


vlad-ivanov-name:
doesn't seem to do anything

PR #1820, opened 2026-02-14
Speed up CI

vlad-ivanov-name:


vlad-ivanov-name:
doesn't seem effective on the runners we get

PR #1823, opened 2026-02-15
Change note cache to custom implementation

christian-schilling:
The git2 API forces a commit for every added note which is very inefficient. It also uses a "dynamic fanout" which causes additional potential overhead.
The custom implementation batches writes to much fewer commits and uses fixed fanout.


christian-schilling:
Still need a test...

josh-core/src/cache/notes.rs
```diff
@@ -11,18 +16,70 @@ impl NotesCacheBackend {
...
+        for (rp, m) in new_entries.into_iter() {
+            if m.len() < FLUSH_AFTER {
+                continue;
+            }
+            for (from, to) in m {
+                let blob = repo.blob(to.to_string().as_bytes())?;
+                builder.upsert(fanout(from), blob, git2::FileMode::Blob.into());
+            }
+            let tree = if let Ok(r) = repo.revparse_single(&rp) {
+                r.peel_to_tree().unwrap()
```
vlad-ivanov-name: .expect() or context()?

josh-core/src/cache/notes.rs
```diff
@@ -55,20 +120,35 @@ impl CacheBackend for NotesCacheBackend {
...
             return Ok(None);
         }
 
-        let key = filter.id();
+        let guard = self.new_entries.lock().unwrap();
+
+        // See if this is one of the newly added entries first
+        let rp = ref_path(filter.id(), sequence_number);
+        if let Some(shard) = guard.get(&rp) {
+            if let Some(to) = shard.get(&from) {
```
vlad-ivanov-name: nit: since we already have modern rust version, you can use && syntax

josh-mq/src/remote.rs
vlad-ivanov-name, line 1: accidentally added?

josh-core/src/cache/notes.rs
christian-schilling: just propagating.

PR #1829, opened 2026-02-18
Add base refs to stacked pushes

christian-schilling:
Add base refs to stacked pushes

In addition to branches for the changes themselves, this creates
branches to be used as the base for PRs so they have clean single
commits diffs.


vlad-ivanov-name:
This shouldn't be in josh-core, creating follow up issue:

https://github.com/josh-project/josh/issues/1835

PR #1901, opened 2026-03-25
Make link files location independent

christian-schilling:
Make link files location independent

Before the path to the .link.josh file would imply
part of the filter (the prefix) but with this change
link files contain the full filter and their location
does not matter anymore.

Change: link-any-place

### DIFF
--- a/josh-cli/src/commands/link.rs
+++ b/josh-cli/src/commands/link.rs
@@ -394,10 +394,8 @@ fn handle_link_push(
         .unwrap_or_else(|| "HEAD".to_string());
 
     // Build the export filter: subdir extracts the link path, :export strips .link.josh
-    let path_filter = josh_core::filter::Filter::new().subdir(normalized_path);
-    let combined_filter = path_filter
-        .export()?
-        .chain(josh_core::filter::invert(*link_file)?);
+    let path_filter = josh_core::filter::Filter::new();
+    let combined_filter = path_filter.chain(josh_core::filter::invert(*link_file)?);
 
     // Apply the filter to get the commit suitable for pushing
     let exported_commit = josh_core::filter_commit(transaction, combined_filter, head_commit.id())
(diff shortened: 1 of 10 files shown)


josh-cli/src/commands/link.rs
```diff
@@ -394,9 +394,9 @@ fn handle_link_push(
         .unwrap_or_else(|| "HEAD".to_string());
 
     // Build the export filter: subdir extracts the link path, :export strips .link.josh
-    let path_filter = josh_core::filter::Filter::new().subdir(normalized_path);
+    let path_filter = josh_core::filter::Filter::new();
     let combined_filter = path_filter
-        .export()?
+        /* .export() */
```
vlad-ivanov-name: should be removed as opposed to commented out?

josh-core/src/filter/mod.rs
```diff
@@ -1034,12 +1034,13 @@ pub fn apply_to_commit2(
         }
         Op::Embed(path) => {
             check_experimental_features_enabled("embed filter")?;
-            let subdir = to_filter(Op::Subdir(path.clone()));
-            let unapply = to_filter(Op::Unapply(LazyRef::Resolved(commit.id()), subdir));
 
             /* dbg!("embed"); */
             /* dbg!(&path); */
             if let Some(link) = read_josh_link(repo, &commit.tree()?, &path, ".link.josh") {
+                //let subdir = to_filter(Op::Subdir(path.clone()));
```
vlad-ivanov-name: same commented out code?

josh-core/src/filter/mod.rs
```diff
@@ -1034,12 +1034,13 @@ pub fn apply_to_commit2(
...
+                let subdir = filter::invert(link.peel())?;
+                let unapply = to_filter(Op::Unapply(LazyRef::Resolved(commit.id()), subdir));
                 /* dbg!(&link); */
```
vlad-ivanov-name: dbg! statements -- not from this PR but can be removed?

vlad-ivanov-name (approved):

PR #1920, opened 2026-04-01
Add blob insertion filter

christian-schilling:
Add blob insertion filter

The new :$path="content" filter inserts a blob with literal string
content at the given path in the filtered tree.

Change: blob-filter


christian-schilling:
Squashed

PR #1944, opened 2026-04-04
Clean up debug output and add label support

christian-schilling:
Clean up debug output and add label support

- Remove debug git ls-tree dumps and DEP: prefix prints from josh-run-container
- Remove -v (verbose file listing) from tar extraction in josh-run
- Add :="..." blob reading in josh-run and josh-run-container; fall
  back to SAFE_NAME when no label is present
- Use [] prefix on all runtime messages for clear attribution
- Replace bare "Image exists/does not exist locally" in josh-run-build-image
  with [image:<short-sha>] prefixed messages
- Add :$$label entries to all workspace definitions

Change: josh-run-labels


christian-schilling:
squashed

PR #1977, opened 2026-04-27
Add support for explicit requires in change footers

christian-schilling:
Add support for explicit requires in change footers

Change: explicit-requires

### DIFF
--- a/josh-changes/src/lib.rs
+++ b/josh-changes/src/lib.rs
@@ -122,6 +122,25 @@ pub fn downstack(
     let change_commit = commits.pop().unwrap();
     let change_parent = change_commit.parent(0)?;
 
+    // Parse Requires: footers, keeping only those referencing changes
+    // actually present in the intermediates
+    let required_raw: std::collections::HashSet<String> = josh_core::get_change_id(&change_commit)
+        .requires
+        .into_iter()
+        .collect();
+    let intermediate_ids: Vec<Option<String>> = commits
+        .iter()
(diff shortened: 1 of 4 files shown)


josh-core/src/lib.rs
```diff
@@ -29,6 +29,7 @@ pub mod submodules;
 pub struct Change {
     pub author: String,
     pub id: Option<String>,
+    pub requires: Vec<String>,
     pub commit: git2::Oid,
```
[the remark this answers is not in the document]
christian-schilling: we don't use semver

josh-changes/src/lib.rs
```diff
@@ -122,6 +122,25 @@ pub fn downstack(
     let change_commit = commits.pop().unwrap();
     let change_parent = change_commit.parent(0)?;
 
+    // Parse Requires: footers, keeping only those referencing changes
+    // actually present in the intermediates
+    let required_raw: std::collections::HashSet<String> = josh_core::get_change_id(&change_commit)
```
vlad-ivanov-name, line 127: would it be better to use BTreeSet here for predictable ordering? might only matter for debug / test prints but still

vlad-ivanov-name (approved):
one comment about btreeset / hashset

josh-changes/src/lib.rs
christian-schilling: We don't print them (or rely on order) anywhere so I think hashset is fine.

PR #1979, opened 2026-04-29
Move Change struct to josh-changes

christian-schilling:
Move Change struct to josh-changes

The struct is only used in that crate.

Change: mv-change

### DIFF
--- a/josh-changes/src/lib.rs
+++ b/josh-changes/src/lib.rs
@@ -1,5 +1,41 @@
 use anyhow::anyhow;
-use josh_core::Change;
+
+#[derive(Debug)]
+struct Change {
+    author: String,
+    id: Option<String>,
+    requires: Vec<String>,
+    commit: git2::Oid,
+}
+
+impl Change {
+    fn new(commit: git2::Oid) -> Self {
+        Self {
+            author: Default::default(),
+            id: Default::default(),
+            requires: Default::default(),
+            commit,
+        }
+    }
+}
+
+fn get_change_id(commit: &git2::Commit) -> Change {
+    let mut change = Change::new(commit.id());
+    change.author = commit.author().email().unwrap_or("").to_string();
+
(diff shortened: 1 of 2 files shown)


josh-changes/src/lib.rs
```diff
@@ -1,5 +1,44 @@
...
+        }
+    }
+}
+
+fn get_change_id(commit: &git2::Commit) -> Change {
+    let mut change = Change::new(commit.id());
+    change.author = commit.author().email().unwrap_or("").to_string();
+
+    let mut have_change_id = false;
+    for line in commit.message().unwrap_or("").split('\n') {
```
vlad-ivanov-name: lines() instead of split()?

josh-changes/src/lib.rs
```diff
@@ -1,5 +1,44 @@
...
+        if !have_change_id && line.starts_with("Change: ") {
```
vlad-ivanov-name: if let Some() = line.strip_prefix()

josh-changes/src/lib.rs
```diff
@@ -1,5 +1,44 @@
...
+            change.id = Some(line.replacen("Change: ", "", 1));
+            // If there is a "Change-Id" as well, it will take precedence
+        }
+        if !have_change_id && line.starts_with("Change-Id: ") {
```
vlad-ivanov-name: same here

josh-changes/src/lib.rs
christian-schilling: fixed

josh-changes/src/lib.rs
christian-schilling: fixed

josh-changes/src/lib.rs
christian-schilling: fixed

vlad-ivanov-name (approved):

PR #1984, opened 2026-04-30
josh-cli: Fix GitHub style (git@github.com:...) URLs

### LINKED ISSUE #1983: Clone error: Failed to resolve path `git@github.com:<...>`
Per [docs](https://josh-project.github.io/josh/reference/cli.html?highlight=ssh#josh-clone), I should be able to `josh clone` an SSH-accessible repo. However,

```bash
$ josh clone git@github.com:josh-project/josh.git :/ josh
Error: Failed to resolve path git@github.com:josh-project/josh.git
Failed to resolve path git@github.com:josh-project/josh.git
No such file or directory (os error 2)
$ josh --version
josh ba5ea29f
```

(`ba5ea29f` is of course the recently tagged r26.04.19 release)

What am I missing?

nstiurca:
Fix #1983

### DIFF
--- a/docs/src/reference/cli.md
+++ b/docs/src/reference/cli.md
@@ -21,7 +21,7 @@ josh clone <url> <filter> <out> [options]
 
 | Argument | Description |
 |----------|-------------|
-| `<url>`  | Remote repository URL (HTTPS, SSH, or local path) |
+| `<url>`  | Remote repository URL — full URL with `http://`, `https://`, `ssh://` or `file://` preferred. |
 | `<filter>` | [Filter spec](./filters.md) to apply (e.g. `:/docs`, `:workspace=workspaces/myproject`) |
 | `<out>`  | Local directory to clone into |
 
@@ -37,13 +37,28 @@ josh clone <url> <filter> <out> [options]
 
 ```shell
 # Clone only the docs/ subdirectory
-josh clone https://github.com/josh-project/josh.git :/docs ./josh-docs
+josh clone http://github.com/josh-project/josh.git :/docs ./josh-docs
 
 # Clone a workspace projection
-josh clone https://github.com/myorg/monorepo.git :workspace=workspaces/frontend ./frontend
+josh clone ssh://github.com/myorg/monorepo.git :workspace=workspaces/frontend ./frontend
 
 # Clone the full repository (no filter)
 josh clone https://github.com/josh-project/josh.git :/ ./josh
+
+# Clone from a mounted network share
+josh clone file:///mnt/network-share/repo :/ ./repo
+```
+
+There is limited support for omitting the URL schema when cloning local paths (`path/to/repo`)
+and scp-style paths (`git@github.com:user/repo.git`). Local paths are canonicalized to absolute paths,
+and scp-style paths are converted to SSH URLs.
+
+```shell
+# Clone using local path
+josh clone ../path/to/repo :/ ./repo
+
+# Clone using scp-style path
+josh clone git@github.com:myorg/monorepo.git :/ ./monorepo
 ```
 
 ---
--- a/josh-cli/src/bin/josh.rs
+++ b/josh-cli/src/bin/josh.rs
@@ -242,14 +242,125 @@ fn to_absolute_remote_url(url: &str) -> anyhow::Result<String> {
         || url.starts_with("file://")
     {
         Ok(url.to_owned())
+    } else if let Some(url) = scp_style_link_to_absolute_remote_url(url) {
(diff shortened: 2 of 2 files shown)


vlad-ivanov-name:
Okay sorry about this CI failure, I gotta fix it on master.

josh-cli/src/bin/josh.rs
```diff
@@ -240,6 +240,7 @@ fn to_absolute_remote_url(url: &str) -> anyhow::Result<String> {
         || url.starts_with("https://")
         || url.starts_with("ssh://")
         || url.starts_with("file://")
+        || url.starts_with("git@")
```
christian-schilling: As the function name says, the output should be an URL. `git@...` is not an URL, so it needs to be converted so a proper url `ssh://...` here.

nstiurca: I updated the logic to convert it to ssh:// URLs. The heuristic is pretty basic (just look for a : without any preceding /). But this happens after the checks for `url.starts_with(...)` so if the heuristic doesn't work, using a proper URL should still work fine.

docs/src/reference/cli.md
```diff
@@ -37,13 +37,28 @@ josh clone <url> <filter> <out> [options]
 
 ```shell
 # Clone only the docs/ subdirectory
-josh clone https://github.com/josh-project/josh.git :/docs ./josh-docs
+josh clone http://github.com/josh-project/josh.git :/docs ./josh-docs
```
christian-schilling, line 40: I don't think this should be changed from https to http

docs/src/reference/cli.md
```diff
@@ -37,13 +37,28 @@ josh clone <url> <filter> <out> [options]
...
 
 # Clone a workspace projection
-josh clone https://github.com/myorg/monorepo.git :workspace=workspaces/frontend ./frontend
+josh clone ssh://github.com/myorg/monorepo.git :workspace=workspaces/frontend ./frontend
```
christian-schilling, line 43: This should stay https as well.

docs/src/reference/cli.md
```diff
@@ -37,13 +37,28 @@ josh clone <url> <filter> <out> [options]
...
 
 # Clone the full repository (no filter)
 josh clone https://github.com/josh-project/josh.git :/ ./josh
+
+# Clone from a mounted network share
```
christian-schilling, line 48: This being a network share is not really relevant or?

docs/src/reference/cli.md
```diff
@@ -37,13 +37,28 @@ josh clone <url> <filter> <out> [options]
...
+josh clone file:///mnt/network-share/repo :/ ./repo
+```
+
+There is limited support for omitting the URL schema when cloning local paths (`path/to/repo`)
```
christian-schilling, line 52: I would not say it's "limited support"...it does work after all. The normalisation to urls for storage is more an implementation detail, do we need to mention that here?

josh-cli/src/bin/josh.rs
```diff
@@ -242,14 +242,125 @@ fn to_absolute_remote_url(url: &str) -> anyhow::Result<String> {
...
+
+/// Try to convert an scp-style URL (user@host:path) to an SSH URL (ssh://user@host/path)
+//
+// This could be an scp-style SSH path, eg git@github.com:user/repo.git
+// it could also be some web path like example.com/path/foo:bar/repo.git
+// or funky file path like dir/path:with:colons/for/some/reason
+//
+// Heuristic: split on first : and verify it looks like a host:path pair.
+fn scp_style_link_to_absolute_remote_url(url: &str) -> Option<String> {
+    if let Some((host, path)) = url.split_once(':') {
```
christian-schilling, line 260: regex? could be more reliable.

docs/src/reference/cli.md
```diff
@@ -21,7 +21,7 @@ josh clone <url> <filter> <out> [options]
 
 | Argument | Description |
 |----------|-------------|
-| `<url>`  | Remote repository URL (HTTPS, SSH, or local path) |
+| `<url>`  | Remote repository URL — full URL with `http://`, `https://`, `ssh://` or `file://` preferred. |
```
christian-schilling, line 24: "prefered" seems like a strange wording. I'd maybe just say "starting with http://, ...."

josh-cli/src/bin/josh.rs
christian-schilling: Thanks @nstiurca, looks mostly good already. I left a few comments on some details.

PR #1986, opened 2026-05-05
Support --base in CLI

vlad-ivanov-name:
Change-Id: I05ec5626182056817151505cd7eca9e94959f002

### DIFF
--- a/tests/cli/push.t
+++ b/tests/cli/push.t
@@ -61,11 +61,11 @@ Push the change back
   To file://${TESTTMP}/remote
      14ecb7c..33f0c00  33f0c009c43980ba5e76995b53f9615a4d880a08 -> master
   
-  Pushed 33f0c009c43980ba5e76995b53f9615a4d880a08 to origin/master
+  Pushed 33f0c009c43980ba5e76995b53f9615a4d880a08 to origin/refs/heads/master
   $ josh push
   Everything up-to-date
   
-  Pushed 33f0c009c43980ba5e76995b53f9615a4d880a08 to origin/master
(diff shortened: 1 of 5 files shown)


tests/cli/push.t
```diff
@@ -77,3 +77,45 @@ Verify the change was pushed to the original repository
...
+  $ echo "feature content" > file3
+  $ git add file3
+  $ git commit -q -m "add file3"
+  $ josh push origin HEAD:refs/heads/feature --base=master
+  To file://${TESTTMP}/remote
+  * (glob)
+  
+  Pushed * to origin/feature (glob)
+
+Verify the new branch on the remote is a descendant of master in the unfiltered repo.
```
christian-schilling: Even though in this case it is, the new branch does not always have to be a descendant of the base branch. If the branch is based on a previous rev in base and base has moved on since, it won't be.

tests/cli/push.t
```diff
@@ -77,3 +77,45 @@ Verify the change was pushed to the original repository
   * 115b269 add file1
   $ cat sub1/file1
   modified content
+
+Pre-create the feature branch on the remote so the underlying git push
+can resolve the unqualified destination ref name (josh push currently
+emits "<sha>:feature" rather than "<sha>:refs/heads/feature").
```
christian-schilling: What? `josh push` does not use the full qualified destination ref? We should fix that or?

christian-schilling (approved):

tests/cli/push.t
vlad-ivanov-name: removed

tests/cli/push.t
vlad-ivanov-name: in PushMode::Normal. changed

PR #2006, opened 2026-05-13
Cache history graph root reachability data

vlad-ivanov-name:
this enables fast reachability checks at the expense of losing some information. the merge_base_many function gives the nearest common ancestor for a set of tips; however, we aren't actually interested in which common ancestor is the nearest one, we only care whether such ancestors exist at all.

Change-Id: I6543d46863e1b7a22955c76708db16e821882c6e

### DIFF
--- a/josh-core/src/cache/transaction.rs
+++ b/josh-core/src/cache/transaction.rs
@@ -1,28 +1,10 @@
+use super::history_graph::compute_sequence_number;
 use super::sled::sled_open_josh_trees;
 use super::stack::CacheStack;
 use anyhow::anyhow;
 
 use std::collections::HashMap;
 use std::sync::{LazyLock, RwLock};
-
-pub const CACHE_VERSION: u64 = 27;
-
-pub trait CacheBackend: Send + Sync {
-    fn read(
-        &self,
-        filter: crate::filter::Filter,
-        from: git2::Oid,
-        sequence_number: u128,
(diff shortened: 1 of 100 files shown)


josh-core/src/cache/transaction.rs
```diff
@@ -452,73 +456,197 @@ pub fn u128_from_oid(oid: git2::Oid) -> u128 {
...
+
+    for c in walk {
+        let oid = c?;
+        let c_commit = transaction.repo().find_commit(oid)?;
+        let parent_infos: Vec<HistoryGraphInfo> = c_commit
+            .parent_ids()
+            .map(|p| collect_history_graph_info(transaction, p))
+            .collect::<anyhow::Result<Vec<_>>>()?;
+        let info = combine_from_parents(oid, &parent_infos);
+        store_history_graph_info(transaction, oid, &info)?;
```
christian-schilling: This will ghost write to the odb every time, which is slower than writing to sled.
I thinks it would be better for non-orphans to just take the roots blob OID from the parents (if there is just one parent or both parents share the same roots blob oid) without reading the blob into vec and re-writing it.

christian-schilling: I'd also consider storing the roots list in sled and not the odb at all.

vlad-ivanov-name: reworked

christian-schilling (approved):

PR #2013, opened 2026-05-14
Use "josh compose run" for CI

christian-schilling:
Use "josh compose run" for CI

Change: josh-run-ci
Change-Series: compose-ci


vlad-ivanov-name (approved):
approving but please disable cache for forks

PR #2022, opened 2026-05-22
Replace diff-applies downstack algorithm with path intersection

christian-schilling:
Replace diff-applies downstack algorithm with path intersection

The old algorithm walked intermediates forward and kept every commit
until the change's diff applied cleanly (and all Requires: were
satisfied). Because it was a single forward pass, it included commits
that were order-independent of the change — any intermediate before
the first clean-apply point was pulled in, even if it touched
completely unrelated files.

The new algorithm does a single backward scan: it seeds an
affected-path set from the change's own modified paths, then marks
only those intermediates whose touched paths intersect that set
(expanding the set as it goes). This lets intermediates that don't
conflict with the change be skipped regardless of their position in
the stack.

As part of this, the Requires: commit footer is replaced by
Change-Series: — authors can assign two or more commits to a named
series to ensure they are merged in order even when they touch
disjoint files. Series membership is encoded as a synthetic
"\x00series:<name>" path entry so it participates in the same
intersection logic without special-casing.

Change: path-intersection-deps

Assisted-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/josh-changes/src/lib.rs
+++ b/josh-changes/src/lib.rs
@@ -4,7 +4,7 @@ use anyhow::anyhow;
 struct Change {
     pub author: String,
     pub id: Option<String>,
-    pub requires: Vec<String>,
+    pub series: Vec<String>,
     pub commit: git2::Oid,
 }
 
@@ -13,7 +13,7 @@ impl Change {
         Self {
             author: Default::default(),
             id: Default::default(),
-            requires: Default::default(),
+            series: Default::default(),
             commit,
         }
     }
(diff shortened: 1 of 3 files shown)


josh-changes/src/lib.rs
```diff
@@ -131,52 +148,40 @@ fn downstack(
...
-        .iter()
-        .map(|s| s.as_str())
-        .filter(|id| available_ids.contains(id))
-        .collect();
+    // Seed the affected path set with the change's own modified paths, then
+    // walk intermediates backwards, keeping any commit whose paths intersect.
+    let change_meta = get_change_id(&change_commit);
+    let mut affected_paths = changed_paths(repo, &change_commit)?;
+    for s in &change_meta.series {
+        affected_paths.insert(format!("\x00series:{}", s));
```
vlad-ivanov-name, line 156: agreed that this will be replaced w/ enum upstack

vlad-ivanov-name (approved):

PR #2025, opened 2026-05-22
Restore sccache R2 remote cache via sidecar proxy

christian-schilling:
Restore sccache R2 remote cache via sidecar proxy

Build containers run with --network none, so sccache cannot reach Cloudflare R2
directly. This adds a sidecar proxy container (aws-sigv4-proxy) that sits on a
shared internal podman network, has external internet access, and signs S3
requests with R2 credentials on behalf of the build container.

The build container reaches only the proxy (internal network, no internet
routing) and never receives credentials. All env var names are declared in
compose.josh — nothing R2/sccache-specific is hardcoded in Rust. When
credentials are absent (e.g. on forks), the proxy is skipped and the existing
local filesystem sccache cache continues to work.

Assisted-By: Claude Sonnet 4.6
Change: sidecar-sccache
Change-Series: compose-ci

### DIFF
--- a/images/sccache-proxy/entrypoint.sh
+++ b/images/sccache-proxy/entrypoint.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+set -eu
+
+# Extract hostname from the full endpoint URL.
+# e.g. https://xxx.r2.cloudflarestorage.com -> xxx.r2.cloudflarestorage.com
+host=$(echo "$SCCACHE_ENDPOINT" | awk -F/ '{print $3}')
+
+exec aws-sigv4-proxy \
+    --name s3 \
+    --region "${SCCACHE_REGION:-auto}" \
+    --host "$host"
--- a/josh-compose/src/container.rs
+++ b/josh-compose/src/container.rs
@@ -4,14 +4,72 @@ use std::process::Command;
 
 use crate::image;
 use crate::job_cache;
-use crate::meta::{self, OutputMode};
+use crate::meta::{self, NetworkMode, OutputMode};
 use crate::podman::{self, PodmanRunArgs};
+
+const SIDECAR_NETWORK: &str = "josh-sccache-net";
+const SIDECAR_IP_PLACEHOLDER: &str = "{SIDECAR_IP}";
+
+/// Read env var names from `sidecar_passthrough` (keys only) and look each up in the
+/// outer process environment. Returns `None` if any listed variable is absent or empty.
+fn resolve_passthrough(passthrough: &[(String, String)]) -> Option<Vec<(String, String)>> {
+    passthrough
+        .iter()
+        .map(|(name, _)| {
+            let val = std::env::var(name).unwrap_or_default();
+            if val.is_empty() {
+                None
+            } else {
+                Some((name.clone(), val))
+            }
+        })
+        .collect()
+}
+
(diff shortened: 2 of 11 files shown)


images/sccache-proxy/entrypoint.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+set -eu
+
+# Extract hostname from the full endpoint URL.
+# e.g. https://xxx.r2.cloudflarestorage.com -> xxx.r2.cloudflarestorage.com
+host=$(echo "$SCCACHE_ENDPOINT" | awk -F/ '{print $3}')
+
+exec aws-sigv4-proxy \
+    --name s3 \
+    --region "${SCCACHE_REGION:-auto}" \
```
vlad-ivanov-name, line 10: we can just set it to auto, no need for this env var.

josh-compose/src/container.rs
```diff
@@ -4,14 +4,72 @@ use std::process::Command;
...
+
+    podman::run_detached(podman::PodmanRunDetachedArgs {
+        image: image_name,
+        name: container_name.clone(),
+        networks: vec!["bridge".to_string(), SIDECAR_NETWORK.to_string()],
+        env_vars,
+    })?;
+
+    // Give the proxy a moment to start listening.
+    std::thread::sleep(std::time::Duration::from_millis(500));
```
vlad-ivanov-name, line 53: this will be fragile; better option is to run a backoff loop (`backon` crate) that queries some health endpoint

josh-compose/src/container.rs
```diff
@@ -4,14 +4,72 @@ use std::process::Command;
...
+                None
+            } else {
+                Some((name.clone(), val))
+            }
+        })
+        .collect()
+}
+
+/// Start the sccache proxy sidecar. Returns the container name.
+fn start_sccache_proxy(
```
vlad-ivanov-name, line 30: i wonder if this could be generalized somehow?

josh-compose/src/podman.rs
```diff
@@ -289,3 +295,100 @@ pub fn run(args: PodmanRunArgs) -> anyhow::Result<RunOutput> {
...
+
+pub struct PodmanRunDetachedArgs {
+    pub image: String,
+    pub name: String,
+    pub networks: Vec<String>,
+    pub env_vars: Vec<(String, String)>,
+}
+
+pub fn run_detached(args: PodmanRunDetachedArgs) -> anyhow::Result<()> {
+    let mut cmd = Command::new("podman");
```
vlad-ivanov-name, line 336: CLI is fragile usually, there's no API?

images/sccache-proxy/entrypoint.sh
christian-schilling: fixed in #2027

josh-compose/src/container.rs
christian-schilling: fixed in #2026

josh-compose/src/container.rs
christian-schilling: fix in #2028

josh-compose/src/podman.rs
christian-schilling: API does not support everything we need (volume import/export) and using it would drag in extra dependencies.

vlad-ivanov-name (approved):

PR #2043, opened 2026-05-26
Store stable diff metadata via store_diff_data in josh-changes

christian-schilling:
Store stable diff metadata via store_diff_data in josh-changes

Add store_diff_data(repo, &Change) that writes a raw commit object
(header + message) to refs/josh/changes at
diffs/<change-id>/<diff-id>/<hash>. Uses raw_header() and
message_raw() to preserve all commit fields. The write is skipped if
the blob already exists. Also factors write_changes_tree helper
shared with write_comment.

Change: store-diff-data

### DIFF
--- a/josh-changes/src/lib.rs
+++ b/josh-changes/src/lib.rs
@@ -487,23 +487,99 @@ pub fn write_comment(
     let diff_id = diff_id(repo, change.commit())?;
 
     let content = serde_json::json!({"message": message}).to_string();
-    let comment_hash =
+    let content_hash =
         git2::Oid::hash_object(git2::ObjectType::Blob, content.as_bytes())?.to_string();
     let blob_oid = repo.blob(content.as_bytes())?;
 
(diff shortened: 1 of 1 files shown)


josh-changes/src/lib.rs
```diff
@@ -487,23 +487,99 @@ pub fn write_comment(
...
+        .id()
+        .ok_or_else(|| anyhow::anyhow!("commit {} has no Change-Id", change.commit()))?;
+    let diff_id = diff_id(repo, change.commit())?;
+
+    let commit = repo.find_commit(change.commit())?;
+    let first_parent = commit
+        .parent_ids()
+        .next()
+        .map(|o| o.to_string())
+        .unwrap_or_else(|| "0000000000000000000000000000000000000000".to_string());
```
vlad-ivanov-name: git2::oid::zero().to_string

vlad-ivanov-name (approved):

PR #2073, opened 2026-05-26
Add prev/next file navigation to josh-gui diff view

christian-schilling:
Add prev/next file navigation to josh-gui diff view

Next/prev file buttons at the top right of the file diff view allow
quick navigation between changed files within the same commit.

Change: josh-gui-file-diff-nav


vlad-ivanov-name (approved):

vlad-ivanov-name:
The manual index iteration could probably be replaced with windows() or something out of itertools

PR #2082, opened 2026-05-26
Add breadcrumb header to all josh-gui pages

christian-schilling:
Add breadcrumb header to all josh-gui pages

Shows repo name (~ for home) on the list page, appends change
subject on detail/diff pages, and filename on file diff pages,
separated by ›.

Change: josh-gui-breadcrumb-header

### DIFF
--- a/josh-gui/src/main.rs
+++ b/josh-gui/src/main.rs
@@ -1,3 +1,5 @@
+use std::sync::OnceLock;
+
 use dioxus::prelude::*;
 
 fn main() {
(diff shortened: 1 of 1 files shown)


josh-gui/src/main.rs
```diff
@@ -54,6 +48,54 @@ fn app() -> Element {
...
+    if let Ok(home) = std::env::var("HOME") {
+        if dir.starts_with(&home) {
+            return dir.replacen(&home, "~", 1);
+        }
+    }
+    dir
+}
+
+fn subject_for_sha(sha: &str) -> String {
+    let repo = match git2::Repository::discover(".") {
```
vlad-ivanov-name: is there no repo anywhere else available in the state? quite hacky. ideally this should be: collect info, then render

josh-gui/src/main.rs
```diff
@@ -54,6 +48,54 @@ fn app() -> Element {
     }
 }
 
+fn repo_name() -> String {
+    let dir = std::env::current_dir()
+        .map(|p| p.display().to_string())
+        .unwrap_or_default();
+    if let Ok(home) = std::env::var("HOME") {
```
vlad-ivanov-name: std::env::home_dir?

vlad-ivanov-name: and you can wrap it in Once

vlad-ivanov-name (approved):

PR #2131, opened 2026-06-05
Split cargo test into build + run to drop git from rust-dev

christian-schilling:
Split cargo test into build + run to drop git from rust-dev

After the per-language image split (d069a9d8d), `rust-dev` still
carried alpine's `git` + `git-daemon` packages solely because one
unit test — `josh-proxy::serve::tests::test_git_cap_discovery` —
shells out to `git http-backend` to verify capability discovery.
No other rust unit/integration test in the workspace spawns a git
child process and no `build.rs` invokes git. Fetch, rust build,
and fmt never need git either.

Split the existing single cargo-test workspace into two phases:

- `ws/cargo-test-build` runs `cargo test --workspace --offline
  --locked --no-run --message-format=json` in `rust-dev-local`,
  extracts every `executable` path under `/debug/deps/` from the
  JSON-lines output, copies them to `/out/test-bins/`. The path
  filter excludes example binaries (which cargo also builds and
  reports under `executable` but lives in `/debug/examples/` and
  would panic when run without their expected env vars).
- `ws/cargo-test-run` consumes `cargo-test-build` as a read-only
  input, iterates `/test-build/test-bins/*`, executes each binary,
  accumulates failures, and exits non-zero if any failed. Runs in
  a new dedicated `rust-test-runner-local` image.

The new `rust-test-runner` image is minimal alpine 3.22.0 + bash,
ca-certificates, git, git-daemon (the alpine package containing
`git-http-backend`), and the runtime libs the cargo binaries link
against (libgcc, libcurl, openssl, zlib). Both `rust-dev` and
`rust-test-runner` are alpine 3.22.0 so library ABIs match across
the build/run boundary.

With the test runner gone, `rust-dev`'s `apk add` line loses `git`
and `git-daemon`. A git package bump no longer invalidates the
rust-dev image used by fetch, build, and fmt.

The orchestrator needed no changes — `ws/test.josh` was already
declaring inputs via `:+ws/<file>` references, so swapping the
inline `:#cargo-test[...]` block for `:+ws/cargo-test-run` pulls
the build phase in transitively just like `build-rust` pulls in
`fetch`.

Verified with `josh compose run`: cargo-test-build copies the
expected binaries (one per crate test target), cargo-test-run
executes them all green including `test_git_cap_discovery` (which
exercises the runner image's `git http-backend`), and the full
pipeline through prysk tests passes.

Change: split-cargo-test-build-run

Assisted-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/ws/cargo-test-build.sh
+++ b/ws/cargo-test-build.sh
@@ -1,0 +1,14 @@
+set -e
+
+cargo test --workspace --offline --locked --no-run --message-format=json \
+    > /tmp/cargo.json
+
+sh check-sccache.sh
+
+mkdir -p /out/test-bins
+
+jq -r 'select(.executable != null) | .executable | select(contains("/debug/deps/"))' \
+    /tmp/cargo.json \
+    | while read -r bin; do
+        cp "$bin" /out/test-bins/
+    done
--- a/images/rust-dev/Dockerfile
+++ b/images/rust-dev/Dockerfile
@@ -5,8 +5,6 @@ FROM ${BASE_IMAGE}
 
 RUN apk add --no-cache \
     bash \
-    git \
-    git-daemon \
     jq \
     zlib-static \
     zlib-dev \
--- a/images/rust-test-runner-local.josh
+++ b/images/rust-test-runner-local.josh
@@ -1,0 +1,4 @@
+bases = :[
+    :#BASE_IMAGE[:+images/rust-test-runner]
+]
+context = :/images/rust-test-runner-local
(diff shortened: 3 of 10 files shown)


ws/cargo-test-build.sh
```diff
@@ -0,0 +1,12 @@
+set -e
+
+cargo test --workspace --offline --locked --no-run --message-format=json \
+    > /tmp/cargo.json
+
+mkdir -p /out/test-bins
+
+sed -n 's/.*"executable":"\([^"]*\)".*/\1/p' /tmp/cargo.json \
```
vlad-ivanov-name: LLMs love to use sed, but it's pretty much unreadable.

there are two options here:

1) use jq -- available in alpine, small and very intuitive query syntax
2) switch to nextest, which has this feature supported natively. the tests will also build and run faster, and there would be only one binary to move around.

https://nexte.st/docs/ci-features/archiving/

christian-schilling: fixed with jq

vlad-ivanov-name (approved):

PR #2139, opened 2026-06-06
Add git tree debugging tools

vlad-ivanov-name:
Add git tree debugging tools

Import git-tree-viewer and git-tree-trace under debug/ as a separate
Cargo workspace (with its own Cargo.lock), excluded from the main
workspace alongside josh-gui.

git-tree-viewer is an egui/eframe desktop app for visualizing git
repository trees, with a Browse mode for inspecting arbitrary revisions
and a Trace mode that receives live tree snapshots. git-tree-trace is
the companion library that merge-queue code uses to send those
snapshots to the viewer.

Co-Authored-By: DeepSeek v4 Pro <noreply@deepseek.com>
Change-Id: I2d0916eefa9ee03a0d403f8d6054b8eed8517191


christian-schilling (commented):
debug/ feels like it could be an output build dir. how about devtools/ ?

vlad-ivanov-name:
sure 👍

christian-schilling (approved):

PR #2142, opened 2026-06-07
Bundle publish/push refs into a single git push

vlad-ivanov-name:
Bundle publish/push refs into a single git push

Previously push and publish ran one `git push` process per ref: an outer
loop over refspec arguments and an inner loop over each computed PushRef.
Since every ref shares one remote URL and a uniform set of flags, they can
all go out in a single `git push <url> <oid1>:<ref1> <oid2>:<ref2> ...`
invocation. This also makes `--atomic` meaningful across the whole set
rather than applying to one ref at a time.

Flatten the prepared pushes into one bundled set, deduplicating by
destination ref name (keep first) so duplicate or colliding refspec
arguments do not make an atomic push reject a repeated destination.

Refactor the surrounding structure to match: rename `run_push` to
`orchestrate_push`, split GitHub PR creation out of the old `execute_push`
into `create_prs`, and reduce the git push itself to `push_refs`. Output
now prints one `Pushing <oid> to <remote>/<ref>` line per ref followed by a
`Pushed N ref(s) to <remote>` summary gated on success.

Assisted-By: claude-opus-4-8
Change: bundle-push-refspecs
Change-Id: Ia6772925cb6f786d7ff6ecfabe6b3c3a30664df3

### DIFF
--- a/josh-cli/src/commands/push.rs
+++ b/josh-cli/src/commands/push.rs
@@ -76,9 +76,7 @@ pub struct PublishArgs {
 }
 
 struct PreparedPush {
-    remote_name: String,
     to_push: Vec<PushRef>,
-    push_mode: PushMode,
     pr_infos: Vec<josh_github_changes::PrInfo>,
 }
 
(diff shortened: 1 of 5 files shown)


christian-schilling:
We might actually not want --atomic here: the push fails for any PR that is already in the gh merge queue. So in a high throughput case where there is always something in the queue this would mean publish can't succeed

vlad-ivanov-name:
Yes, but the atomic flag is passed by user, and not enabled by default. Or do you mean atomic should be removed altogether?

christian-schilling (approved):

PR #2157, opened 2026-06-11
Polish josh-gui list selection, scrolling, and comment styling

christian-schilling:
Polish josh-gui list selection, scrolling, and comment styling

Make a row click select the row instead of navigating, and move the
detail-page navigation onto the Subject cell click (with stopPropagation
so it doesn't fall back to the row handler). Capture the scroll
container's MountedData in onmounted and restore the scroll position
from scroll_offset right there, so navigating back from the detail view
returns to where the user was scrolled. Guard the selected-change effect
with a last-seen-selection signal so it only scrolls the selection into
view when the selection actually changes after mount, and use the
mounted element's typed scroll() API for that instead of a JS eval.

Restyle the selected row as light-on-dark (#d8d8d8 bg, #1a1a1a text) and
override the per-cell color classes (review/vote/check status, num,
loading) so they stay readable against the inverted background. Give
related rows a blue tint (#22303f / #2d4054) so they read as
contextually-linked rather than just hovered. Bump comment bodies to a
0.95rem system font in #e8a872 so they're easier to scan, and drop the
unused .diff-comment-inline rules.

Change: gui-list-selection-polish
Assisted-By: anthropic/claude-opus-4-7

### DIFF
--- a/josh-gui/src/list.rs
+++ b/josh-gui/src/list.rs
@@ -82,8 +82,17 @@ pub fn ListView(
         });
     }
 
+    let mut scroll_el: Signal<Option<std::rc::Rc<MountedData>>> = use_signal(|| None);
+
+    // Only react to selection changes made after mount; on mount the scroll
+    // position is restored from scroll_offset in onmounted instead.
+    let mut last_sel = use_signal(|| selected_change.peek().clone());
     use_effect(move || {
         let sel = selected_change.read();
+        if *last_sel.peek() == *sel {
+            return;
+        }
+        last_sel.set(sel.clone());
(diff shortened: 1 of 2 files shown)


josh-gui/src/list.rs
```diff
@@ -82,8 +82,24 @@ pub fn ListView(
         });
     }
 
+    use_effect(move || {
+        let offset = *scroll_offset.peek();
+        if offset > 0 {
+            let js = format!(
```
vlad-ivanov-name: this is super hacky and i would expect it to have side effects

vlad-ivanov-name (approved):

PR #2175, opened 2026-06-13
Avoid history walks with pins and workspaces used together

vlad-ivanov-name:
Avoid history walks with pins and workspaces used together

Previously, use of :pin filter would trigger extra history walks when
stored filters change per revision. This could happen for example with
workspaces, stored filters, or hooks.

This happened because the optimizer would not be able to look through
the :pin and figure out that the subtract could be collapsed.

However, for this particular place in code where extra parents would be
synthesized, we are not concerned with the actual :pin semantics at all,
since on a tree level they are a noop and are handled otherwise
elsewhere.

Therefore, before computing the subtracted tree, legalize pin away, so
that the Subtract works properly.

Change-Id: Ib8ababe434e08a3b5cbe5d31cd3ebbd325aec3ab


christian-schilling (approved):
josh vote: approve

PR #2199, opened 2026-06-18
Intern filter Ops into hash-consed nodes

christian-schilling:
Intern filter Ops into hash-consed nodes

The filter optimizer paid two repeated costs per node. to_op() locked a
global map and deep-cloned the stored Op (the Vec<Filter> of a large
Compose/Chain, plus paths) on each of its ~37 hot call sites, and
to_filter() re-ran build_op() -- a gix tree serialize plus SHA-1 -- on
every call, even though the great majority rebuilt an Op whose value was
already known.

Intern each Op into a single leaked &'static Node that owns the Op and
caches its OID in a OnceLock, so build_op() runs at most once per unique
Op. to_op_ref() hands out a borrow into that node instead of cloning, and
the optimizer reads through &Op, cloning only the inner Vec at the few
sites that mutate it. Filter stays content-addressed by its OID, so
persistence, the distributed/sled/transaction caches, sentinels and
pretty-printed output are all unchanged.

The interner is keyed by the Op itself through a structural Hash/Eq
rather than a hand-built byte fingerprint; a small Regex newtype (compared
by pattern string) lets Op derive those traits. This also deduplicates the
regex/squash/rev variants that the old fingerprint skipped.

Change: intern-filter-ops

### DIFF
--- a/josh-filter/src/op.rs
+++ b/josh-filter/src/op.rs
@@ -1,11 +1,37 @@
 use crate::filter::Filter;
 use anyhow::anyhow;
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub enum LinkMode {
     Embedded,
     Snapshot,
     Pointer,
+}
+
+/// Newtype around `regex::Regex` adding structural `PartialEq`/`Eq`/`Hash` (by pattern
+/// string) so `Op` can derive them for use as an interning key. Derefs to the inner regex.
+#[derive(Clone, Debug)]
+pub struct Regex(pub regex::Regex);
+
+impl std::ops::Deref for Regex {
+    type Target = regex::Regex;
+    fn deref(&self) -> &regex::Regex {
+        &self.0
+    }
+}
+
+impl PartialEq for Regex {
+    fn eq(&self, other: &Self) -> bool {
+        self.0.as_str() == other.0.as_str()
+    }
+}
+
+impl Eq for Regex {}
+
+impl std::hash::Hash for Regex {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.0.as_str().hash(state);
+    }
 }
 
 impl std::fmt::Display for LinkMode {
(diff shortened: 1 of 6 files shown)


josh-filter/src/op.rs
```diff
@@ -1,13 +1,39 @@
...
     Snapshot,
     Pointer,
 }
 
+/// Newtype around `regex::Regex` adding structural `PartialEq`/`Eq`/`Hash` (by pattern
+/// string) so `Op` can derive them for use as an interning key. Derefs to the inner regex.
+#[derive(Clone, Debug)]
+pub struct Regex(pub regex::Regex);
+
+impl std::ops::Deref for Regex {
```
vlad-ivanov-name, line 16: for this and below: we already have derive_more in ws dependencies

josh-filter/src/persist.rs
```diff
@@ -501,14 +532,32 @@ impl InMemoryBuilder {
...
-    let oid = git2::Oid::from_bytes(tree_id.as_bytes()).unwrap();
+// Canonicalize an `Op` to its single interned node via structural equality. On a hit we
+// return the existing node and drop `op`; on a miss we leak a node (its `Op` lives for the
+// process, exactly like the old `FILTERS` map) and key it by a clone of the `Op`.
+fn intern(op: Op) -> &'static Node {
+    let mut intern = INTERN.lock().unwrap();
+    if let Some(node) = intern.get(&op) {
+        return node;
+    }
+    let node: &'static Node = Box::leak(Box::new(Node {
```
vlad-ivanov-name, line 543: can you try with Arc instead and see if it produces any meaningful perf difference?

josh-filter/src/persist.rs
```diff
@@ -501,14 +532,32 @@ impl InMemoryBuilder {
...
-    let f = Filter(oid);
-    FILTERS.lock().unwrap().entry(f).or_insert(op);
-    f
+pub fn to_filter(op: Op) -> Filter {
+    let node = intern(op);
+    let oid = *node.oid.get_or_init(|| {
+        let mut builder = InMemoryBuilder::new();
+        let tree_id = builder.build_op(&node.op).expect("failed to build op");
+        let oid = git2::Oid::from_bytes(tree_id.as_bytes()).unwrap();
+        FILTERS.lock().unwrap().entry(Filter(oid)).or_insert(node);
```
vlad-ivanov-name, line 557: would it make sense to move this out of the get_or_init lock region?

vlad-ivanov-name (approved):

josh-filter/src/persist.rs
vlad-ivanov-name: Agreed to switch to Arc upstack

PR #2200, opened 2026-06-18
Make Filter a node handle with a lazy content OID

christian-schilling:
Make Filter a node handle with a lazy content OID

The optimizer rebuilds large intermediate Compose/Chain ops on every step,
and each rebuild paid for content hashing: to_filter() forced build_op() --
a gix tree serialize plus SHA-1 -- eagerly on every call, and to_op_ref()
locked a global map and looked up the node on every child access. Almost all
of those ops are thrown away before any OID is ever needed.

Make Filter a pointer to the interned &'static Node rather than the content
Oid. Interning already canonicalizes one node per structurally-unique Op, so
filter identity is node identity: equality and hashing become pointer ops and
to_op_ref() is a plain field borrow with no lock. The content Oid is computed
lazily in Filter::id() (the only caller of build_op() now), so it is built
once per filter that is actually persisted or used as a cache key -- never for
the optimizer's throwaway intermediates. Filter::id() stays byte-identical, so
persistence and the per-filter caches are unchanged.

Sentinels (sequence_number / reachable_roots) become memoized leaked nodes
with their Oid pre-seeded, keeping their !=/id() behaviour under pointer
identity. as_tree() walks the reachable node set instead of dumping the global
filter map, and the now-unread Oid->node reverse map is removed.
PassthroughHasher gains a write_usize path for the pointer-keyed maps.

This is about a 5x speedup on the ultrawide_pin benchmark.

Change: filter-node-handle

### DIFF
--- a/josh-filter/src/persist.rs
+++ b/josh-filter/src/persist.rs
@@ -2,34 +2,25 @@ use anyhow::{Context, anyhow};
 use gix_object::WriteTo;
 use gix_object::bstr::BString;
 use std::collections::HashMap;
-use std::hash::BuildHasherDefault;
 use std::sync::{LazyLock, OnceLock};
 
-use crate::filter::{Filter, reachable_roots, sequence_number};
-use crate::hash::PassthroughHasher;
+use crate::filter::Filter;
 use crate::op::{BlobContent, LazyRef, Op, Regex, RevMatch};
 
-/// An interned, immutable `Op` together with its lazily-computed `Filter` OID.
-/// Nodes are leaked (`&'static`) and live for the process lifetime, exactly like the
-/// `Op`s previously stored in `FILTERS`. The OID is built at most once via `build_op`.
+/// An interned, immutable `Op` together with its lazily-computed content OID.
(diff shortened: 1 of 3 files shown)


josh-filter/src/persist.rs
```diff
@@ -49,29 +40,28 @@ pub fn to_op(filter: Filter) -> Op {
...
-/// Get the known `Filter` -> `Op` mappings for use in as_tree/from_tree.
-pub fn get_filters() -> HashMap<Filter, Op, BuildHasherDefault<PassthroughHasher>> {
-    FILTERS
-        .lock()
-        .unwrap()
-        .iter()
-        .map(|(f, node)| (*f, node.op.clone()))
-        .collect()
+/// Enumerate the direct child filters of an op, for reachable-set traversal in `as_tree`.
+fn child_filters(op: &Op) -> Vec<Filter> {
```
vlad-ivanov-name, line 51: kinda looks like it should be in impl Op?

christian-schilling: Why?

vlad-ivanov-name (approved):

PR #2201, opened 2026-06-18
Pack filtered objects via an in-memory ODB backend

christian-schilling:
Pack filtered objects via an in-memory ODB backend

josh wrote every object it produced while filtering as a loose object: one
file, one zlib stream and one directory insert each, which dominates the cost
of large rewrites. Route those writes through a process-global in-memory
libgit2 ODB backend instead, and flush them to packfiles.

The backend is a custom git_odb_backend over an scc::HashMap, registered at
high priority on every repository opened through a Transaction, so existing
repo.blob()/treebuilder/commit calls land in memory with no filter call-site
changes. Reads check memory first and fall through to the on-disk pack/loose
backends on a miss. git2 does not expose the raw odb/repository pointers, so
register() reads the pointer out of the single-field Repository newtype; this
is sound only while git2 is pinned exactly and must be re-verified on upgrade.

Objects are flushed to a packfile (OIDs sorted first, so the pack name is
deterministic) and evicted at the boundaries where an external git process
reads them from disk: spawn_git_command (CLI), the proxy's write_to_repo
(before git http-backend serves) and push_head_url (before pushing upstream).
A drop-time flush keyed on a live-transaction counter is the backstop for
pure josh-filter runs. Proxy serve/push and CLI clone/push/fetch all stay
correct because the in-memory backend is invisible to those subprocesses
until the boundary flush runs.

The .t snapshots that list .git/objects change from loose objects to
packfiles accordingly; no command output changed.

Left to do:
- Memory is unbounded within a single operation: there is no threshold or
  background mid-operation flush yet, so a very large rewrite holds all new
  objects in RAM until the boundary flush. This was deferred from the plan.
- Flush evicts, so re-reads of an object after its flush hit the on-disk
  pack rather than RAM; a memory budget or lazier eviction could help.
- read_prefix is unimplemented, so abbreviated OIDs are not resolved from the
  in-memory store (full-OID lookups, the hot path, are unaffected).
- DistributedCacheBackend and josh-filter's GitNotesFilterHook open their own
  repositories without the backend, so they cannot see in-flight filtered
  objects; the distributed cache is consequently sparser for chained filters.
- The flush at every external-git boundary is correct but coarse; a more
  scoped store could avoid flushing unrelated objects in the proxy.

Change: in-memory-odb-backend

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3029,6 +3029,7 @@ dependencies = [
  "josh-filter",
  "josh-starlark",
  "josh-test-support",
+ "libgit2-sys",
  "log",
  "percent-encoding",
  "pest",
@@ -3038,6 +3039,7 @@ dependencies = [
  "regex",
  "rs_tracing",
  "rustc-hash",
+ "scc",
  "serde",
  "serde_json",
  "serde_yaml",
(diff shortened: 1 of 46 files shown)


vlad-ivanov-name:
two things for this PR:

1) have an agent look at the git2-rs fork with backend trait impl
2) use that as a reference and extract backend into a new crate

vlad-ivanov-name:
Superseded by other PR

PR #2217, opened 2026-07-03
Pack filtered objects on a background thread

christian-schilling:
Pack filtered objects on a background thread

Writing a packfile -- zlib-compressing every buffered object and fsyncing the
result -- is the expensive tail of a MemOdb flush. Running it inline blocked the
filter hot path, most of all the mid-transaction overflow flush that fires
repeatedly during a large rewrite.

Funnel all packing to one process-global background worker (josh-git-data's new
flusher module). A store's overflow now enqueues a best-effort chunk and returns
immediately, so the write path keeps filtering while the pack is built; a
chunk_in_flight guard collapses the burst of overflowing writes that follow into
a single queued chunk. The transaction- and external-git-boundary flush enqueues
a drain and blocks until the objects are durable. A single worker processes jobs
FIFO, so a store's queued chunks always complete before its drain and no two
packbuilders ever run against the same store concurrently.

Stores stay per-operation, so a job carries the store's Arc and its repo path (a
git2::Repository is not Send); the worker opens its own handle to run the
packbuilder. Because the write path and the flusher now touch a store from
different threads, pack_to_disk evicts only the oids it snapshotted -- writes
that land during a background chunk stay buffered for the next chunk or the
drain -- rather than clearing the whole map.

Boundary drains remain synchronous, so on-disk output is unchanged: overflow
does not trigger at test-suite object volumes, and no snapshot moves.

Change: mem-odb-background-flusher
Assisted-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/josh-git-data/src/flusher.rs
+++ b/josh-git-data/src/flusher.rs
@@ -1,0 +1,98 @@
+//! Background packfile writer for [`MemOdb`](crate::mem_odb::MemOdb).
+//!
+//! Writing a packfile (zlib-compressing every buffered object and fsyncing the result) is the
+//! expensive tail of a flush. Running it inline would block the filter hot path — the mid-run
+//! overflow flush most of all, which fires repeatedly during a large rewrite. So all packing is
+//! funnelled to one process-global worker thread:
+//!
+//! * [`enqueue_chunk`] hands the worker a store to pack and returns immediately. It is best-effort
+//!   (fire-and-forget), used from the write path when a store overflows its size limit.
+//! * [`drain`] hands the worker a store and blocks until it is packed and evicted, used at
+//!   transaction and external-git boundaries where the objects must be durable on disk before the
+//!   caller proceeds.
+//!
+//! A single worker processes jobs FIFO, so a store's queued overflow chunks always complete before
+//! its drain — and no two packbuilders ever run against the same store concurrently. Stores are
+//! per-operation, so a job carries the store's `Arc` and its repo path (a `git2::Repository` is not
+//! `Send`); the worker opens its own repository handle to run the packbuilder.
+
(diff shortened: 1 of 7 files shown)


josh-git-data/src/flusher.rs
```diff
@@ -0,0 +1,98 @@
...
+/// A unit of packing work for the background worker. Each carries the store's `Arc` (the packbuilder
+/// reads objects back out of it) rather than a repository handle, which is not `Send`.
+enum Job {
+    /// Pack the store's currently-buffered objects and evict them. Best-effort; no acknowledgement.
+    Chunk { store: Arc<MemOdb> },
+    /// Pack and evict, then acknowledge, so a boundary caller can block until the objects are on
+    /// disk.
+    Drain {
+        store: Arc<MemOdb>,
+        ack: SyncSender<Result<(), String>>,
```
vlad-ivanov-name, line 34: anyhow::Error should be possible to use here. or anyhow::Result for the whole thing.

josh-git-data/src/flusher.rs
```diff
@@ -0,0 +1,98 @@
...
+        .send(Job::Drain { store, ack: ack_tx })
+        .is_err()
+    {
+        return Err(git2::Error::from_str(
+            "mem-odb flusher channel disconnected",
+        ));
+    }
+    match ack_rx.recv() {
+        Ok(Ok(())) => Ok(()),
+        Ok(Err(msg)) => Err(git2::Error::from_str(&msg)),
```
vlad-ivanov-name, line 93: esp. here we essentially are using git2::Error as a type erased error type so makes sense to just switch to anyhow

vlad-ivanov-name (approved):

PR #2218, opened 2026-07-03
Skip futile trailing-Compose distribution in flatten (experimental)

christian-schilling:
Skip futile trailing-Compose distribution in flatten (experimental)

flatten distributes Chain[.., Compose(Z)] into Compose([Chain[.., z_j]]). When
the Compose is the trailing element with a non-empty prefix, step's common_pre
re-merges the result straight back to Chain[prefix, Compose(Z)], so the
distribution is a round-trip that converges to the input.

In ultrawide_pin this shape is reached via the pin-legalization Subtract in
per_rev_filter: legalize_pin exposes compose(pinned) as a bare trailing element
of each of the N wide-chains, and flatten distributing it builds an
O(N*|pinned|) tree that step then collapses again. Skipping that case turns that
path from O(N*|pinned|) into O(N). Leading/middle Compose elements still
distribute; their branches are not re-merged.

Experimental, and deliberately not behaviour-neutral: for some
subtract/exclude-of-compose filters the distribution had enabled a downstream
collapse, so skipping it leaves a valid but less-collapsed representation. The
pretty_print, filter_id and workspace_errors snapshots are updated to that form.
Refining the guard to skip only the truly-futile Chain[prefix, compose(pinned)]
round-trip -- without giving up the subtract/exclude collapse -- is future work.

Change: skip-trailing-compose-distribution
Assisted-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/josh-filter/src/opt/flatten.rs
+++ b/josh-filter/src/opt/flatten.rs
@@ -53,6 +53,22 @@ pub fn flatten(filter: Filter) -> Filter {
                     if !others_invertible {
                         break;
                     }
+                    // EXPERIMENTAL: distributing a trailing Compose that has a non-empty prefix is
+                    // often futile: it produces Compose([Chain[prefix, z_j]]) which step()'s
+                    // common_pre re-merges back to Chain[prefix, Compose(Z)]. For the
+                    // Chain[file_chain, compose(pinned)] shape (reached via pin-legalization in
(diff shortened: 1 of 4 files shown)


christian-schilling:
Closed in favor of #2223

PR #2265, opened 2026-07-18
Add views rfc

christian-schilling:
Add views rfc

This is sort of a big feature so we should think it through

Change: views-rfc

### DIFF
--- a/rfcs/views.md
+++ b/rfcs/views.md
@@ -1,0 +1,299 @@
+# RFC: Views
+
+## Summary
+
+Introduce *views*: named, versioned filter objects stored as git refs. A view is any ref whose
+commit tree contains a `view.josh` file. Views supersede the current workspace mechanism: the
+workspace behavior (compose a root directory with its in-tree stored filter) becomes the default
+template instantiated on the view mechanism. The existing workspace feature and views coexist for
+a migration period; deprecating and eventually removing `Op::Workspace` in favor of the nearly
+identical `Op::Stored` is a future step, taken only after users have had time to migrate. Views
+give a versioned, mutable home for per-workspace policy (meta options, compat flags) and for
+migration points (`:rev`), without rewriting source history.
+
+## Motivation
+
+The current workspace design entangles the filter definition with the history it filters:
+`workspace.josh` lives inside the thing it defines. This entanglement is the root cause of several
+of josh's hairiest mechanisms and limitations:
+
+- **No home for policy.** Meta options (`:~(...)` history flags, gpgsig handling, future compat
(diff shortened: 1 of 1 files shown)


rfcs/views.md
```diff
@@ -0,0 +1,299 @@
+# RFC: Views
+
+## Summary
+
+Introduce *views*: named, versioned filter objects stored as git refs. A view is any ref whose
+commit tree contains a `view.josh` file. Views supersede the current workspace mechanism: the
```
vlad-ivanov-name, line 6: at the root or where?

rfcs/views.md
```diff
@@ -0,0 +1,299 @@
...
+migration points (`:rev`), without rewriting source history.
+
+## Motivation
+
+The current workspace design entangles the filter definition with the history it filters:
+`workspace.josh` lives inside the thing it defines. This entanglement is the root cause of several
+of josh's hairiest mechanisms and limitations:
+
+- **No home for policy.** Meta options (`:~(...)` history flags, gpgsig handling, future compat
+  flags such as a pattern-semantics flag) have no per-workspace, versioned, mutable place to live.
```
vlad-ivanov-name, line 21: i don't think "mutable" makes sense in this context

rfcs/views.md
christian-schilling: See below...

rfcs/views.md
christian-schilling: It kind of does. That's exactly the difference from stored filters: For those "views" only current (mutable) state of the filter matters, not it's history.

rfcs/views.md
```diff
@@ -0,0 +1,299 @@
...
+leading position, not an element of the filter spec. Ordinary filter chaining *after* the view
+reference is allowed and well-defined: resolve the view to filter F (with its pragmas), then chain
+the remaining ops on top of F's output.
+
+Consequence, accepted deliberately: views cannot reference other views (no view op means no
+`:=bar` inside a `view.josh`). For v1 this structurally rules out cross-ref resolution cycles. It
+is a one-way door that can be opened later (promote to an op with cycle detection) but never
+closed.
+
+### The workspace template and filter/code co-evolution
```
vlad-ivanov-name, line 149: we should define a way to pin what the view is applied to. so we have two things that are potentially pinned.

christian-schilling: I'm not sure what you mean by this. But in any case. The rfc can/should still be refined after being merged

vlad-ivanov-name (approved):

PR #2267, opened 2026-07-18
Warn when filter files silently degrade to :empty and bump cache version

christian-schilling:
Warn when filter files silently degrade to :empty and bump cache version

Filter files that fail to load have so far been silently replaced by :empty, which makes whole
parts of a composed filter disappear without any indication of what went wrong. This happened in
practice when the "blob" to "insert" rename of the starlark Filter method broke
ws/tests-per-category.star: starlark evaluation failed, get_starlark() logged only at trace level
and returned :empty, and "josh compose run" silently dropped all per-category prysk test jobs.

Make all of these fallbacks loud:

- get_starlark(): log a warning (instead of a trace message) when starlark evaluation fails, and
  also when applying the subfilter fails, including the script path and the error.
- get_filter(): log warnings when parsing the filter file fails, when legalize_stored() fails, and
  when the resulting filter is not invertible. The fallback behavior itself is unchanged.

To avoid spurious noise when filtering a full history that contains commits with a broken script
that is already fixed at the requested commit, the starlark warnings are only emitted when the
failing script is part of the root tree of a commit explicitly requested via filter_commit()
(tracked in a process-global set of requested tree ids), and only once per unique script blob.
Failures in other (historical) trees are still logged at trace level.

The josh and josh-filter CLIs initialized env_logger with its default filter, which only shows
error-level messages unless RUST_LOG is set, so these warnings would have stayed invisible.
Initialize env_logger with a default filter of "error,josh_core::filter=warn" so the new fallback
warnings are shown by default (on stderr) while RUST_LOG still overrides the filter entirely. The
default is deliberately limited to the josh_core::filter module: all fallback warnings live there,
and a broader josh_core-wide default would also surface unrelated pre-existing warnings (e.g.
"filter_refs: Can't filter reference" and "walk: can't hide") in normal use and in test output.

Also bump CACHE_VERSION from 28 to 29: commit fe1daddb changed the invert semantics of generative
filters (Insert/TreeId) from Exclude to Empty without bumping the cache version, so version-28
sled and distributed caches hold stale pre-change apply/unapply results and warm caches make new
binaries return the old, wrong compose trees.

Update the prysk test expectations accordingly: cache refs and cache directory listings now show
version 29 instead of 28, and filter/workspace_trailing_slash.t now expects the new parse warning
for its deliberately invalid workspace file.

Change: loud-filter-fallbacks
Assisted-By: Claude Fable 5 <noreply@anthropic.com>


christian-schilling:
Will be fixed in a better way (hard errors) in the future.

PR #2269, opened 2026-07-18
Benches: prepare rust-lang filtering bench

vlad-ivanov-name:
Benches: prepare rust-lang filtering bench

For now, only using josh-filter.

Change-Id: I62e1b3de4d2ac789aaa56c545c0e7271a8b96fd4


christian-schilling:
I'm wondering: Why even make a history with "injected" workspace? All the "era" filters can be put directly into the :rev(...).
Also if still going the intermediate "inject" route: Why workspace (which hopefully will be deprecated in favor of stored filter) instead of plain stored filter?

vlad-ivanov-name:
>I'm wondering: Why even make a history with "injected" workspace? 

mostly to demonstrate a hypothetical scenario of how it would work if rust was using stored filters during its development in the past. rev-ranges and insertion are also an additional demo of features.

>Why workspace

switched to plain stored filters.

christian-schilling (approved):

PR #2282, opened 2026-07-19
Avoid using libgit2 built-in cache in favor of custom LRU

vlad-ivanov-name:
Avoid using libgit2 built-in cache in favor of custom LRU

Change-Id: Ie2bcd0e97503fe28af2e27ffce7e846ada41110a


christian-schilling (dismissed):

vlad-ivanov-name:
Turns out the inner libgit2 cache is more complex than assumed; this solution is suboptimal

PR #2292, opened 2026-07-20
Pass object data around as slice-arcs instead of vecs

vlad-ivanov-name:
Pass object data around as slice-arcs instead of vecs

Git object data is immutable; it will not be modified or resized.
Therefore, it would not only work with a boxed slice, but can also work
with Arcs -- different places that use object data can just store arcs.

Change-Id: I05916d8278764d488804783d06747188a601574c


vlad-ivanov-name:
The approach itself is valid but will need to be reused in a different way

PR #2298, opened 2026-07-22
Memoize filtered parent tree ids in create_filtered_commit2

christian-schilling:
Memoize filtered parent tree ids in create_filtered_commit2

create_filtered_commit2 loaded every filtered parent commit via find_commit only to read its
tree id and its own oid, and select_parent_commits/rewrite_commit used nothing else from them.
Carry the parents as (oid, tree_id) pairs where the tree id comes from a per-transaction memo
populated whenever josh writes a filtered commit, so a parent produced earlier in the same walk
is not re-parsed from the odb. rewrite_commit now takes parent oids directly (it only formatted
their ids anyway), so no parent commit is loaded on the common linear path.

The saving is invisible with libgit2's object cache on -- the reload is a cache hit -- but that
cache masks the real cost: with it disabled (which approximates the contended-lock regime under
concurrent proxy load), filtering deep history improves 4.1% at 1000 commits and 2.1% at 10000
on deephistory_subdir (p<0.05). Full josh compose run suite passes.

Change: filtered-parent-tree-id-memo
Assisted-By: Claude Opus 4.8 (claude-opus-4-8)


christian-schilling:
closed in favor of #2301

PR #2309, opened 2026-07-23
Avoid redundant work in remove_pred

christian-schilling:
Avoid redundant work in remove_pred

The Op::Pattern arm rebuilt unchanged trees on every apply. In tree::remove_pred:
- return the result oid instead of a Tree, so warm glob-cache hits skip repo.find_tree entirely;
  the single caller re-reads the tree only when filtering actually changed it and returns the
  input Rewrite untouched otherwise
- take the predicate as &dyn Fn(&str, bool) and reuse one String path buffer instead of
  allocating a PathBuf per entry, matching via Pattern::matches_with (identical to
  matches_path_with on &str)
- hoist empty_id() out of the entry loop and drop the dead pred(&path, false) subtree shortcut
- track whether any entry changed and return the input oid without writing a new tree when
  nothing did (guarded by a debug_assert that the rebuilt tree is bit-identical)

Also drop a Rewrite clone by copying the tree oid before calling remove_pred.

Filtered results are bit-identical; new unit tests cover gitlink/symlink filemodes, legacy-mode
and entry-order normalization, and the unchanged fast path. Criterion benches, measured for
this change together with pattern-glob-in-op below it against the pre-change base (on the
variant that still memoized the compiled glob in a global map -- the split forms only remove
per-apply work): deephistory_glob 10k commits +26.3%/+31.8%/+20.3% throughput
(recursive/prefix/literal), widetree_glob up to +44.4% at 50000 entries (prefix), incremental
recursive cases -24% to -43% time; no stable case regressed.

Change: pattern-arm-alloc-fixes

### DIFF
--- a/josh-core/src/filter/mod.rs
+++ b/josh-core/src/filter/mod.rs
@@ -1368,13 +1368,19 @@ pub fn apply<'a>(
                 require_literal_separator: true,
                 require_literal_leading_dot: true,
             };
-            Ok(x.clone().with_tree(tree::remove_pred(
+            let input = x.tree().id();
+            let t = tree::remove_pred(
                 transaction,
-                "",
-                x.tree().id(),
-                &|path, isblob| isblob && (glob.matches_path_with(path, options)),
+                &mut String::new(),
+                input,
+                &|path, isblob| isblob && glob.matches_with(path, options),
                 peel_filter(filter).id(),
-            )?))
+            )?;
+            Ok(if t == input {
+                x
+            } else {
+                x.with_tree(repo.find_tree(t)?)
+            })
         }
         Op::Insert(dest_path, content) => {
             let (oid, mode, is_tree) = match content {
--- a/josh-core/src/filter/tree.rs
+++ b/josh-core/src/filter/tree.rs
@@ -84,51 +84,120 @@ pub fn regex_replace<'a>(
     Ok(repo.find_tree(builder.write()?)?)
 }
 
-pub fn remove_pred<'a>(
-    transaction: &'a cache::Transaction,
-    root: &str,
+/// Compare two tree entry names in canonical git tree order: byte-wise, with tree entries
+/// sorted as if their name had a trailing '/'.
+fn git_tree_entry_cmp(a: &[u8], a_is_tree: bool, b: &[u8], b_is_tree: bool) -> std::cmp::Ordering {
+    let len = a.len().min(b.len());
+    match a[..len].cmp(&b[..len]) {
+        std::cmp::Ordering::Equal => {}
+        ord => return ord,
+    }
+    let ca = a
+        .get(len)
+        .copied()
+        .unwrap_or(if a_is_tree { b'/' } else { 0 });
+    let cb = b
+        .get(len)
+        .copied()
+        .unwrap_or(if b_is_tree { b'/' } else { 0 });
+    ca.cmp(&cb)
+}
+
+/// Rebuild `input` keeping only blob entries accepted by `pred`. `path` is a reusable buffer
+/// holding the slash-separated path of the tree currently being visited; it is restored to its
+/// incoming length before returning. Gitlink (submodule) entries are always dropped.
+pub fn remove_pred(
+    transaction: &cache::Transaction,
+    path: &mut String,
     input: git2::Oid,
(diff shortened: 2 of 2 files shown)


josh-core/src/filter/mod.rs
```diff
@@ -1363,19 +1363,48 @@ pub fn apply<'a>(
...
+            // pointers to `Box::leak`'d, immutable nodes that live for the process lifetime, so
+            // entries are pure functions of their key and can never go stale -- this map is
+            // intentionally not wired into `reset_caches`.
+            static PATTERN_MEMO: LazyLock<
+                std::sync::RwLock<
+                    std::collections::HashMap<usize, std::sync::Arc<(glob::Pattern, git2::Oid)>>,
+                >,
+            > = LazyLock::new(Default::default);
+
+            let node_key = peel_op_ref(filter) as *const Op as usize;
```
vlad-ivanov-name: can this casting be moved into whatever type peel_op_ref() returns, or as an extension trait? so it's peel_op_ref(filter).key() or something like this.

reason: right now it reads like a hack and not an established convention -- the intention of casting is not clear; the comment above would need to go to implementation of key(), so this concept is explained in one place and more semantically readable

josh-core/src/filter/tree.rs
```diff
@@ -84,51 +84,113 @@ pub fn regex_replace<'a>(
...
         let name = entry.name().ok_or_else(|| anyhow!("INVALID_FILENAME"))?;
-        let path = std::path::PathBuf::from(root).join(name);
-
-        if entry.kind() == Some(git2::ObjectType::Blob) && pred(&path, true) {
-            builder.insert(name, entry.id(), entry.filemode()).ok();
+        let base = path.len();
+        if !path.is_empty() {
+            path.push('/');
+        }
+        path.push_str(name);
```
vlad-ivanov-name, line 140: this would benefit from being extracted, even if you extract this into a nested function. so something like normalize_path with a comment on why it's defaulting to /?

josh-core/src/filter/tree.rs
```diff
@@ -84,51 +84,113 @@ pub fn regex_replace<'a>(
...
     if let Some(cached) = transaction.get_glob((input, key)) {
-        return Ok(repo.find_tree(cached)?);
+        return Ok(cached);
     }
 
+    let repo = transaction.repo();
     let tree = repo.find_tree(input)?;
     let mut builder = repo.treebuilder(None)?;
+    let empty = empty_id();
+    let mut changed = false;
```
vlad-ivanov-name, line 128: would benefit from a comment about what this "changed" is

josh-core/src/filter/tree.rs
```diff
@@ -84,51 +84,113 @@ pub fn regex_replace<'a>(
...
+                    if builder.insert(name, entry.id(), entry.filemode()).is_err() {
+                        changed = true;
+                    }
+                } else {
+                    changed = true;
+                }
+            }
+            Some(git2::ObjectType::Tree) => {
+                let s = remove_pred(transaction, path, entry.id(), pred, key)?;
+                if s != entry.id() || s == empty || entry.filemode_raw() != 0o0040000 {
```
vlad-ivanov-name, line 172: no enums/constants available for this?

josh-core/src/filter/tree.rs
```diff
@@ -84,51 +84,113 @@ pub fn regex_replace<'a>(
...
+                    changed = true;
+                }
+            }
+            // Gitlinks (and any other kinds) are dropped, so the rebuilt tree differs from the
+            // input and must not take the unchanged fast path below.
+            _ => {
+                changed = true;
             }
         }
+        path.truncate(base);
```
vlad-ivanov-name, line 188: definitely deserves a comment on what's happening here

josh-core/src/filter/tree.rs
christian-schilling: fixed

josh-core/src/filter/tree.rs
christian-schilling: done

josh-core/src/filter/tree.rs
christian-schilling: See added TODOs, we have to revisit this.

vlad-ivanov-name (approved):

PR #2310, opened 2026-07-23
Walk glob patterns component-wise with subtree pruning

christian-schilling:
Walk glob patterns component-wise with subtree pruning

Op::Pattern previously matched every blob's full path against the whole glob via
tree::remove_pred, recursing into every subtree regardless of whether the pattern could still
match anything below it. Replace that walk with tree::remove_pattern: the pattern is split into
'/'-separated components, each component is compiled once, and the walk advances an NFA state
bitmask of active component positions per tree level. Subtrees where no position can match
anything below are pruned without recursion, and a trailing all-'**' suffix short-circuits
per-entry matching to a dotfile check, preserving require_literal_leading_dot semantics.

The component split works on the glob parser's own token stream instead of re-scanning the
pattern string: the vendored glob crate is patched to expose its parser output (PatternToken
and CharSpecifier become public, plus a Pattern::tokens() accessor and a Pattern::from_tokens()
constructor that compiles a component from a token sub-slice). That makes the split exact by
construction -- a '/' inside a bracket class lives inside the class token and never appears as
a Char('/') separator, and '**' arrives pre-validated as its own path component -- so bracket
classes containing '/' need no fallback: under require_literal_separator a class can never
match a separator, in the full-path walk and the per-component walk alike.

The glob result cache key gains the state mask -- (input tree, pattern key, state) -- so cached
subtree results stay independent of the path they were reached through. Patterns with more than
63 components (the u64 mask limit) fall back to the legacy full-path walk under a synthetic
root-keyed cache entry at state 0. The compiled form is the new
josh_filter::pattern::CompiledPattern, replacing the plain compiled glob as the Op::Pattern
payload (still built at op construction, with Eq/Hash on the source string); the tree walk
stays in tree.rs and takes the cache key as a parameter. A differential unit test compares
remove_pattern against reference full-path glob matching over randomized trees with dotfiles
and collisions, and new prysk tests cover dotfile and path-aliasing behavior.

Criterion, vs the previous commit: deephistory_glob prefix -61%, literal -65%, recursive -9%;
widetree_glob prefix -89% (9x), sparse -25%, recursive -19%; no case regressed.

Change: pattern-nfa-walk

### DIFF
--- a/josh-filter/src/pattern.rs
+++ b/josh-filter/src/pattern.rs
@@ -1,0 +1,154 @@
+use glob::PatternToken;
+
+/// The MatchOptions of the `Op::Pattern` arm: full-path glob matching with literal separators
+/// and literal leading dots. Also exactly the options under which the component-wise walk of
+/// `tree::remove_pattern` is equivalent to a full-path match.
+pub const PATTERN_MATCH_OPTIONS: glob::MatchOptions = glob::MatchOptions {
+    case_sensitive: true,
+    require_literal_separator: true,
+    require_literal_leading_dot: true,
+};
+
+/// One '/'-separated component of a glob pattern.
+#[derive(Clone)]
+pub enum PatternComponent {
+    /// A component that is exactly `**`: matches zero or more non-dot-leading path components.
+    Star2,
+    /// Any other component, matched against single entry names.
+    Glob(glob::Pattern),
+}
+
+/// The payload of `Op::Pattern`: a glob compiled at construction -- use `Op::pattern` -- for
+/// the component-wise NFA walk of `tree::remove_pattern`. `PartialEq`/`Eq`/`Hash` are the
+/// source pattern string (recovered verbatim by `as_str`), so `Op` can derive them for use as
+/// an interning key.
+///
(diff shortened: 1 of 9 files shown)


josh-filter/src/pattern.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+}
+
+/// Split `pattern` at '/' into raw component strings, respecting bracket classes (a '/' inside
+/// `[...]` is a class member, not a separator). Returns `None` for degenerate patterns: a bracket
+/// class containing '/' (which can never match under `require_literal_separator`) or a malformed
+/// class (unreachable after the full pattern compiled). Mirrors the bracket scan of glob-0.3.3:
+/// an optional leading '!', a ']' in first member position is a literal member, then the closing
+/// ']'.
+fn split_pattern_components(pattern: &str) -> Option<Vec<String>> {
```
vlad-ivanov-name: `glob` has an internal AST type but it's not public. let's vendor it and patch.

vlad-ivanov-name (approved):

PR #2333, opened 2026-07-26
Port remove_pred and remove_pattern tree rebuilding to gix-object

christian-schilling:
Port remove_pred and remove_pattern tree rebuilding to gix-object

Parse input trees directly from raw odb bytes with gix_object::TreeRef
and rebuild them as entry vectors serialized by gix, instead of going
through libgit2's treebuilder. Writes go straight to the git2 odb, so
they stay immediately visible to all readers (and land in the in-memory
memodb store when registered) while git2 still owns all I/O.

Byte-identity with the treebuilder path is preserved explicitly:

- tree_entry_name_valid() replicates libgit2's valid_entry_name
  (empty, slash, dot, dot-dot, and the NTFS ".git" aliases; the HFS
  aliases are now rejected on all platforms, where libgit2 only
  checked them on Apple hosts)
- normalize_filemode() replicates libgit2's mode normalization
  including its quirks (any exec bit wins over symlink/gitlink types)
- the TreeRebuild accumulator replicates the name-keyed treebuilder's
  last-insert-wins behavior for duplicate names, scanning only when an
  entry-order violation has been detected among the kept entries, and
  owns the unchanged fast path guarded by its changed flag

deephistory_glob and widetree_glob improve by 5-17% (no libgit2 FFI or
object-cache locking per tree, zero-copy entry iteration); all other
benches are unchanged.

Change: gix-port-remove-pred-pattern

### DIFF
--- a/josh-core/src/filter/tree.rs
+++ b/josh-core/src/filter/tree.rs
@@ -1,4 +1,5 @@
 use super::*;
+use crate::objects;
 use anyhow::anyhow;
 
 pub fn pathstree<'a>(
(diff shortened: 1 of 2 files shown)


josh-core/src/filter/tree.rs
```diff
@@ -103,6 +104,40 @@ fn git_tree_entry_cmp(a: &[u8], a_is_tree: bool, b: &[u8], b_is_tree: bool) -> s
...
+/// round-trip through a filter even when it keeps them entirely. The intended semantics is
+/// byte-preservation: pass fully-kept subtrees through verbatim (an order violation alone must
+/// not set `changed`), and when a rebuild is forced by dropping or replacing entries, keep the
+/// surviving entries in their original relative order, duplicates included -- duplicates share
+/// a name, so a path predicate keeps or drops them together. This needs an order-preserving
+/// alternative to `write_tree_now` and tolerant (non-bisecting) name lookups in
+/// subtract/intersect/overlay, and it changes filtered oids for repos containing such trees,
+/// so it must land as its own change with tests -- not inside the gix port, which is validated
+/// on byte-identity against the libgit2 baseline and keeps this wart until then.
+fn push_deduped(
```
vlad-ivanov-name: shared a refactoring proposal

vlad-ivanov-name (approved):

PR #2336, opened 2026-07-26
Add worktree-centric CLI RFC

christian-schilling:
Add worktree-centric CLI RFC

Replace the clone-based drop-in model with a bare store plus filtered
worktrees: filters attach to worktrees rather than remotes, levels
connect histories via filter edges, and inside a worktree everything
is plain git.

Change: worktree-rfc

### DIFF
--- a/rfcs/worktree.md
+++ b/rfcs/worktree.md
@@ -1,0 +1,321 @@
+# RFC: Worktree-centric CLI
+
+## Summary
+
+Today josh-cli mirrors git's porcelain: `josh clone <url> <filter> <dir>` creates a working
+copy checked out at the filtered projection, and `josh fetch` / `pull` / `push` are
+filter-aware drop-in replacements for their git counterparts, with the filter fixed at clone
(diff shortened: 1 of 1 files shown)


rfcs/worktree.md
```diff
@@ -0,0 +1,321 @@
...
+  only writes back into the filter's image. Filters become guardrails, not just conveniences.
+- **A clean division of labor.** Everything local is plain git and *actually* works, because it
+  operates on real, local, filtered history — no pretend upstreams, no parity chasing. Josh
+  appears only at the seams: creating projections and syncing with upstream.
+
+## Design
+
+### `josh init` creates a store, `josh remote add` attaches upstreams
+
+There is no `josh clone`. Cloning means "make me a copy of *that* repository", which quietly
```
vlad-ivanov-name, line 76: End users might still be more comfortable with josh clone aliased to init and/or remote commands

vlad-ivanov-name (approved):

PR #2355, opened 2026-07-30
Rework josh pull into quiet, change-aware josh changes pull

vlad-ivanov-name:
Rework josh pull into quiet, change-aware josh changes pull

`josh pull` used to shell out to `git pull` with stdio inherited, which
floods the output with the stacked-changes refs (`@changes/`/`@base/`/
`@heads/`) carried by the namespace remote.

Fetch with `git fetch --porcelain` (requires git >= 2.41) instead, parse
it into a typed `RefUpdate` enum (josh-cli/src/porcelain.rs), and print a
curated summary: one line per updated branch, plus a change count for
the stacked-changes refs.

The command moves from `josh pull` to `josh changes pull` to distinguish
it from `git pull`; `josh fetch` stays top-level. Integration is always
rebase-style (the --rebase/--autostash flags are gone): fast-forward,
or rewind/restack skipping local commits whose Change-Id trailer already
landed upstream (e.g. via a merge queue), cherry-picking the rest on top.
Anything more complex -- merge commits in the local stack, diverged
commits without a Change-Id, restack conflicts -- fails with refs
untouched. Uncommitted changes are stashed before integrating and
re-applied afterwards (autostash).

The "updated N changes" summary line counts a republished change only
when its content actually changed, compared via `git patch-id --stable`,
so restacking without edits does not over-report.

Builds on the StackedRef type and the GitCommand builder from the same
Change-Series; it does not compile without them.

Change: changes-pull-porcelain
Change-Series: changes-pull-porcelain
Assisted-By: kimi-code/k3


josh-core/src/git.rs
```diff
@@ -218,3 +218,53 @@ pub fn read_parent_ids(repo: &git2::Repository, oid: git2::Oid) -> anyhow::Resul
         .map(|p| Ok(git2::Oid::from_bytes(p.as_bytes())?))
         .collect()
 }
+
+/// Spawn a git command with output captured, returning stdout on success.
```
christian-schilling: The means no progress indicator. That's not so great. Can't we capture stdout but inherit stderr?

vlad-ivanov-name: fixed

vlad-ivanov-name: wait no it still doesn't print 🤔 will check

vlad-ivanov-name: okay, checked, it does, but only if there are actual changes -- git prints them if it gets to pack negotiation stage

christian-schilling (approved):

PR #2358, opened 2026-07-31
Support publishing josh changes to Gerrit

christian-schilling:
Support publishing josh changes to Gerrit

Add Gerrit as a forge for `josh changes publish`, selected explicitly with
`--forge gerrit` (Gerrit servers are not identifiable from their URL). Unlike
the GitHub path, publishing to Gerrit is a plain `git push` to the magic
`refs/for/<branch>` ref -- the push itself creates or updates the reviews, so
there is no API call or token and authentication is handled by git.

Because Gerrit keys a change by its Change-Id and expects it to appear exactly
once with a single parent, josh's `downstack` model (where a change can belong
to several stacks with differing ancestry) cannot be mapped one-to-one. Two
per-remote modes, selected with `--gerrit-mode` and stored as a `gerrit-mode`
meta key, resolve this:

- `independent` (default): push only the changes that have no dependencies --
  those sitting directly on the target base -- each as its own separately
  submittable review. A change that still depends on unmerged work is skipped
  until its predecessors merge. Only roots are pushed, so no change is ever
  duplicated across stacks.
- `stack`: push the whole commit history once as a single Gerrit relation
  chain -- one change per commit, each with exactly one parent.

Every pushed commit is rewritten to carry a `Change-Id: I<40 hex>` trailer,
derived deterministically from the change's josh id (and reused as-is if already
valid), so re-publishing lands as a new patchset on the same Gerrit change
rather than a duplicate. The trailer is written only onto the pushed commits;
local history is untouched.

Assisted-By: anthropic/claude-opus-4-8
Change: changes-publish-gerrit

### DIFF
--- a/josh-cli/src/commands/push.rs
+++ b/josh-cli/src/commands/push.rs
@@ -4,7 +4,7 @@ use josh_changes::{PushMode, PushRef, StackedChangeRef, StackedRef, build_to_pus
 use josh_core::git::normalize_repo_path;
 
 use crate::config::{RemoteConfig, read_remote_config};
-use crate::forge::Forge;
+use crate::forge::{Forge, GerritMode};
 use crate::porcelain::PushRefUpdate;
 
 #[derive(Debug, clap::Parser)]
@@ -107,6 +107,7 @@ fn prepare_push(
     filter: josh_core::filter::Filter,
     push_mode: &PushMode,
     forge: &Option<Forge>,
+    gerrit_mode: GerritMode,
     dry_run: bool,
 ) -> anyhow::Result<PreparedPush> {
     let repo = transaction.repo();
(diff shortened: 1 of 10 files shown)


josh-cli/src/commands/push.rs
```diff
@@ -242,11 +264,40 @@ fn push_refs(
...
 ) -> anyhow::Result<()> {
     if to_push.is_empty() {
         return Ok(());
     }
 
+    // Gerrit: every change targets the same magic ref `refs/for/<branch>`, so a
+    // single bundled push (multiple source commits, one destination ref) is not
+    // possible -- push each ref in its own invocation. `refs/for` refs are never
+    // force-pushed.
+    if *forge == Some(Forge::Gerrit) {
```
vlad-ivanov-name: this is quite hacky.

push_refs is a function that exists to support branch-based forges. here we attempt to reconstruct gerrit push of what is essentially a github push?

i think a more clean solution would be to rename this to push_branch_based, and then have another fn like push_change_based; otherwise i feel like those two have little to do with each other, and that's also signified by early return

vlad-ivanov-name (approved):

PR #2381, opened 2026-08-04
Add RFC for replacing the Starlark filter with a WASM filter

christian-schilling:
Add RFC for replacing the Starlark filter with a WASM filter

Proposes replacing the experimental :!path[context] Starlark filter with
a WebAssembly-based scripted filter: modules resolved directly from the
commit tree, parameterized via an =arg,... list in the filter spec,
executed in a deterministic sandboxed runtime (wasmi) with fuel and
memory limits, and memoized by (module OID, args, filtered tree OID).
Starlark compatibility is retained via an interpreter compiled to wasm.

Change: wasm-filter-rfc

### DIFF
--- a/rfcs/wasm-filter.md
+++ b/rfcs/wasm-filter.md
@@ -1,0 +1,468 @@
+# RFC: Replace the Starlark filter with a WASM filter
+
+## Summary
+
+Replace the experimental `:!path[context]` Starlark filter (`josh-starlark` crate)
+with a WebAssembly-based scripted filter. A commit's tree contains a wasm module;
+josh instantiates it in a sandboxed, deterministic runtime, gives it read-only
+access to the (context-filtered) tree, and the module returns the filter to apply.
+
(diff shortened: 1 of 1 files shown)


rfcs/wasm-filter.md
```diff
@@ -0,0 +1,468 @@
...
+Plain core-wasm ABI, no component model, no WASI. Strings are UTF-8
+`(ptr, len)` in guest memory; filters are `u32` handles.
+
+Guest exports:
+
+| Export | Purpose |
+|---|---|
+| `memory` | linear memory |
+| `josh_abi_version: () -> u32` | must return `1`; anything else → error |
+| `josh_alloc: (len: u32) -> ptr` | host allocates guest buffers through this |
```
vlad-ivanov-name, line 216: can't we inherit something more standard 🤔 there's nothing really special about josh allocs

vlad-ivanov-name (approved):
we gotta convert to buck2 before that because it will be a nightmare to build and maintain otherwise

PR #2391, opened 2026-08-05
Release the sled cache lock before the compose container run

christian-schilling:
Release the sled cache lock before the compose container run

josh compose held the process-global sled cache open for its whole run,
so its exclusive file lock blocked every other josh command in the repo
for the entire container build -- even though only the initial workspace
filtering touches the cache; running the containers just reads objects
from the repo.

Add Transaction::release_cache, which flushes and drops every sled
handle (the transaction's own trees, the shared backend's cached trees
via a new CacheBackend::release, and the global db) so sled's lock is
freed. The transaction's repo and in-memory object store stay usable, so
josh compose calls it right after computing the workspace tree and before
the container phase. The cache trees become Option so they can be dropped;
any access after release panics, which is correct since the only caller
is done with the cache.

Change: sled-release-compose-lock

### DIFF
--- a/josh-compose/src/lib.rs
+++ b/josh-compose/src/lib.rs
@@ -58,6 +58,11 @@ pub fn run(
 
     let (ws_tree, _safe_name) = filter::compute_ws_tree(transaction, &filter_spec, source_commit)?;
 
+    // Computing the workspace tree is the only cache-backed work; running the containers below
+    // only reads objects from the repo. Release the cache lock now so its long tail doesn't block
+    // other josh processes from opening the shared sled cache.
+    transaction.release_cache()?;
+
     let mut attempted = std::collections::HashSet::new();
     // Only extract output artifacts into the working tree when running against
     // uncommitted changes (input_ref == "."). For committed refs there is no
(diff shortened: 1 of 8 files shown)


vlad-ivanov-name:
Why does it need to exist beyond the point of the release anyway? Two potential solutions:

* a different type that transaction can be converted into. like if we need to be able to read objects we could do it like in gix -- some other type + shared trait ReadObjects or something like that
* just open a new transaction again from that point with a different cache stack

christian-schilling:
> Why does it need to exist beyond the point of the release anyway? Two potential solutions:
> 
> 
> 
> * a different type that transaction can be converted into. like if we need to be able to read objects we could do it like in gix -- some other type + shared trait ReadObjects or something like that
> 
> * just open a new transaction again from that point with a different cache stack
> 
> 
> 
> 

The problem is that proxy opens sled process wide. So the whole "tie it to instance lifetime" approach does not work.
So I'd say do this "hack" now to fix the workflow pain (not being able to push/pull changes while compose runs) and then refactor this properly down the line. (Sooner rather than later)

vlad-ivanov-name (approved):

PR #2417, opened 2026-08-14
Disable DNS on internal sidecar network; symlink aadvark binaries to avoid base drift

vlad-ivanov-name:
Disable DNS on internal sidecar network; symlink aadvark binaries to avoid base drift

The sidecar attaches to both the internal sidecar network and the
default bridge, so podman writes both networks' aardvark-dns addresses
into resolv.conf in nondeterministic order (Go map iteration over
NetworkStatus). The internal network's aardvark-dns refuses to forward
external names and answers NXDOMAIN, and Go's resolver treats NXDOMAIN
as terminal without trying the second nameserver. Whenever the internal
resolver landed first, the sccache proxy could not resolve the upstream
endpoint and sccache's startup storage check failed the build.

Steps reach sidecars by raw IP, so the internal network does not need
DNS; disabling it makes resolution deterministic via the bridge network.

install-podman.sh symlinked the podman binary and OCI runtimes from the
pinned static bundle, but not the helpers (netavark, aardvark-dns,
conmon, ...), so podman resolved them from the runner image's distro
packages. Those drift as ubuntu-latest moves, changing container
networking/DNS behavior underneath a pinned podman.

Link the bundle's helpers into /usr/local/libexec/podman, which is first
in podman's default helper search path, ahead of the distro locations.

Change: pin-podman-helpers
Assisted-By: moonshotai/kimi-k3

### DIFF
--- a/.github/workflows/scripts/install-podman.sh
+++ b/.github/workflows/scripts/install-podman.sh
@@ -28,6 +28,15 @@ for bin in podman crun runc pasta fuse-overlayfs fusermount3; do
     sudo ln -sf "${PREFIX}/usr/local/bin/${bin}" "/usr/local/bin/${bin}"
 done
 
+# Pin the helper binaries from the bundle too (netavark, aardvark-dns, conmon,
+# ...). /usr/local/libexec/podman is first in podman's default helper search
+# path, so without this podman silently falls back to the runner image's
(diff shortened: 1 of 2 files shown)


vlad-ivanov-name:
Superseded

PR #2481, opened 2026-08-21
Share object stores and defer ref publication

christian-schilling:
Share object stores and defer ref publication

Share each in-memory object store by objects directory so unpublished objects can outlive transactions without producing tiny packs. Queue ref edits and flush objects before publishing refs at transaction boundaries.

Change: pack-lifetime
Assisted-By: anthropic/claude-opus-5
Assisted-By: openrouter/openai/gpt-5.6-sol

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3659,6 +3659,7 @@ dependencies = [
  "gix-zlib",
  "josh-gix-ext",
  "log",
+ "parking_lot 0.12.5",
  "tempfile",
 ]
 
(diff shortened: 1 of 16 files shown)


vlad-ivanov-name (commented):
1) flush_all should probably be FlushGuard with flush on drop; we should document how and when it should be used
2) we should send the refs to updated to be written _after_ object flushing

vlad-ivanov-name (approved):

PR #2512, opened 2026-08-25
Support building and running josh on Windows

### LINKED ISSUE #2235: Failed to compile on Windows
I tried this command on Windows:

```sh
cargo install josh-cli --locked --git https://github.com/josh-project/josh.git --tag r26.06.11
```

and got this error:

```
Compiling josh-compose v26.6.11
error[E0425]: cannot find function `getuid` in crate `libc`
   --> josh-compose\src\container.rs:257:30
    |
257 |     let uid = unsafe { libc::getuid() };
    |                              ^^^^^^
    |
   ::: C:\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\libc-0.2.186\src\windows\mod.rs:517:5
    |
517 |     pub fn getpid() -> c_int;
    |     ------------------------- similarly named function `getpid` defined here
    |
help: a function with a similar name exists
    |
257 -     let uid = unsafe { libc::getuid() };
257 +     let uid = unsafe { libc::getpid() };
    |

error[E0425]: cannot find function `getgid` in crate `libc`
   --> josh-compose\src\container.rs:258:30
    |
258 |     let gid = unsafe { libc::getgid() };
    |                              ^^^^^^
    |
   ::: C:\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\libc-0.2.186\src\windows\mod.rs:517:5
    |
517 |     pub fn getpid() -> c_int;
    |     ------------------------- similarly named function `getpid` defined here
    |
help: a function with a similar name exists
    |
258 -     let gid = unsafe { libc::getgid() };
258 +     let gid = unsafe { libc::getpid() };
    |

error[E0425]: cannot find function `getuid` in crate `libc`
   --> josh-compose\src\image.rs:26:30
    |
 26 |     let uid = unsafe { libc::getuid() };
    |                              ^^^^^^
    |
   ::: C:\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\libc-0.2.186\src\windows\mod.rs:517:5
    |
517 |     pub fn getpid() -> c_int;
    |     ------------------------- similarly named function `getpid` defined here
    |
help: a function with a similar name exists
    |
 26 -     let uid = unsafe { libc::getuid() };
 26 +     let uid = unsafe { libc::getpid() };
    |

error[E0425]: cannot find function `getgid` in crate `libc`
   --> josh-compose\src\image.rs:27:30
    |
 27 |     let gid = unsafe { libc::getgid() };
    |                              ^^^^^^
    |
   ::: C:\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\libc-0.2.186\src\windows\mod.rs:517:5
    |
517 |     pub fn getpid() -> c_int;
    |     ------------------------- similarly named function `getpid` defined here
    |
help: a function with a similar name exists
    |
 27 -     let gid = unsafe { libc::getgid() };
 27 +     let gid = unsafe { libc::getpid() };
    |

For more information about this error, try `rustc --explain E0425`.
error: could not compile `josh-compose` (lib) due to 4 previous errors
warning: build failed, waiting for other jobs to finish...
error: failed to compile `josh-cli v26.6.11
```

The previous version `r26.05.08` compiles fine.

This seems to be introduced in d2d19dc10eedca0dc3b0307fcc663b33541ba070. Would it be possible to gate `josh compose` for unix only, since the podman stuff probably won't work on Windows anyway.

### LINKED ISSUE #2288: josh clone fails with `fatal: '//\\?\Z:\path\to\dir' does not appear to be a git repository`
With `r26.05.08` on Windows, cloning fails after fetching with `fatal: '//\\?\Z:\path\to\dir' does not appear to be a git repository` (where `Z:\path\to\dir` is the absolute path to the relative dir specified with the `josh clone` command (e.g., `Z:\path\to $ josh clone https://github.com/my/awesome-repo :/some-dir ./dir`).

Behaviour is as expected in WSL with Ubuntu for same command.

JohnCampionJr:
Closes #2235. Fixes #2288.

`josh-proxy` and the `josh` / `josh-filter` CLIs now build and run on Windows. Nothing here changes unix behavior — every change is either a `cfg` split whose unix arm is the existing code, or portable-by-construction.

I saw the note on #2235 that you have no infrastructure to test this platform, so the aim was to bring evidence rather than a patch: everything below is validated by functional suites on real Windows machines, x86-64 and ARM64.

## The changes

| Crate | Change |
|---|---|
| `josh-gix-ext` | Add `component_bytes`, the portable form of the path-component → bytes conversion used to match tree entry names (`josh-gix-ext`, `josh-search` and `josh-core` each did this with `std::os::unix::ffi::OsStrExt::as_bytes`). Worth noting: rustc suggests `std::os::windows::prelude::OsStrExt` here, which is **not** equivalent — that trait yields UTF-16 code units, not bytes. |
| `josh-core` | `replace_child_inner` refuses an empty tree entry name rather than writing one git will reject — reachable if a path component has no byte representation on the platform. |
| `josh-rpc` | Gate `tokio_fd` behind `cfg(unix)`: it wraps raw fds with libc for the SSH shell, while `josh-proxy` only uses `josh_rpc::calls` (plain serde types). |
| `josh-proxy` | Split `serve_namespace` — SSH serving relays git's stdio over unix sockets, so non-unix returns "unsupported on this platform" rather than failing to compile. HTTP serving is unaffected. |
| `josh-proxy` | Portable hook installation: symlink on unix (unchanged), `sh` shim elsewhere, since Windows symlinks may require elevated privileges. Hook mode dispatches on `argv[0]`, which a shim can't fake for a native executable, so the shim passes `JOSH_HOOK=<name>` and dispatch honors both. The shim runs an absolute, single-quoted path (git runs hooks with `GIT_DIR` as cwd). Hooks dir via `fs::create_dir_all` rather than shelling to `mkdir`, which isn't an executable on Windows. |
| `josh-proxy` | **Dual-stack listener.** A bare `[::]` bind accepts IPv4 on Linux (`bindv6only` off by default) but is v6-only on Windows: the proxy starts, logs a healthy address, and refuses every client dialing `127.0.0.1`. Now sets `IPV6_V6ONLY=0` explicitly. This one is a cross-platform behavior change (explicit rather than OS-default) — happy to split it into its own PR if you'd prefer. |
| `josh-compose-podman` | `host_uid_gid` returns `1000:1000` on non-unix, mirroring the default user of the VM podman runs containers in. Makes the CLI build; makes no claim about compose on Windows. |
| `josh-cli` | Valid `file://` URLs on Windows (#2288). `canonicalize()` returns extended-length paths that git rejects inside `file://`; `\\?\UNC\server\share` keeps its authority as `file://server/share/...`, `\\?\C:\...` becomes `file:///C:/...`. |

Plus `docs/src/contributing/windows.md`: setup, build, run, testing, and a troubleshooting table of every failure hit getting here.

## Testing

`josh compose run` needs podman, which isn't available on Windows, so the port was validated with two hermetic functional suites driving the built binaries against a local git-HTTP server — no network, fresh temp dirs 

- **proxy**: boot/teardown, filtered clone (`:prefix`), pinned-SHA fetch (`@sha`), reverse-filter push asserting re-rooted parentage, warm-cache reuse across a proxy restart
- **cli**: `josh-filter`, `josh clone` into a relative directory (the #2288 trigger shape), `josh changes pull`, `josh push --base` asserting exact parentage

These now run in CI, in a Windows job added by this PR, on **x86-64 and arm64**. Both also pass on macOS arm64 and Linux, and `cargo check --workspace --all-targets` is clean on unix, so the unix side is unregressed.

## Scope notes

- **`josh-ssh-shell` is untouched and still doesn't build on Windows** — it uses unix sockets, fifos, raw fds and `libc::isatty` directly, independent of anything here, so `cargo build --workspace` doesn't succeed there while `-p josh-proxy -p josh-cli` does. Gating it is mechanical but noisy and orthogonal; happy to do it separately if you want the workspace green on Windows.
- **ARM64 needs clang-cl** for the `aws-lc-sys` dependency (MSVC can't assemble its GNU-syntax ARM assembly). The runner image provides it, so the CI job uses it there.
- Two behaviors I noticed but deliberately did not change, in case they're intentional: `get_head` falls back to `refs/heads/master` when a remote's HEAD is unborn, so a filtered clone comes back empty but "successful"; and `josh push` without `--base` writes a parentless commit where the proxy path refuses outright. Happy to file either as an issue.

Change: windows-support
Assisted-By: anthropic/claude-opus-5


docs/src/contributing/windows.md
```diff
@@ -0,0 +1,134 @@
...
+git clone "http://127.0.0.1:8000/josh-project/josh.git:prefix=josh.git" filtered-test
+```
+
+## Testing
+
+The repository's own test suite runs through `josh compose run`, which needs podman; compose is
+not supported on Windows yet, so the Windows work was validated with two hermetic functional
+suites that drive the built binaries against a local git-HTTP server (no network, fresh temp
+directories):
+[rigsmith/josh-binaries](https://github.com/rigsmith/josh-binaries) — `test/roundtrip.sh`
```
vlad-ivanov-name: would prefer if this was not linked here -- we can't guarantee authenticity of the binaries published there

docs/src/contributing/windows.md
```diff
@@ -0,0 +1,134 @@
...
+```powershell
+winget install Rustlang.Rustup
+winget install Microsoft.VisualStudio.2022.BuildTools --override "--wait --passive --add Microsoft.VisualStudio.Workload.VCTools --includeRecommended"
+winget install Git.Git
+winget install GoLang.Go
+```
+
+Open a **new** terminal afterwards so the tools land on PATH. The repo pins its Rust version
+via `rust-toolchain.toml`, so the first cargo command inside the clone fetches the exact
+toolchain — nothing to select by hand.
```
vlad-ivanov-name: this is probably too detailed to be included here -- can be omitted entirely

docs/src/contributing/windows.md
```diff
@@ -0,0 +1,134 @@
...
+
+Why each is needed:
+
+- **MSVC Build Tools** — rust's Windows target links with `link.exe`, and C dependencies
+  (`libz-sys`, `aws-lc-sys`) compile with `cl.exe` during the build.
+- **Git for Windows** — needed at *runtime*, not just to clone: josh shells out to `git`, the
+  hooks josh installs are `sh` shims run by its bundled sh, and its `bash.exe` runs the test
+  suites.
+- **Go** — builds the suites' local git-HTTP server (they are hermetic; no network).
```
vlad-ivanov-name: could also be skipped. all this section really needs is a list of winget packages

docs/src/contributing/windows.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
+Point the C-compiler variables at it. Persist them in your PowerShell profile so every shell
+has them:
+
+```powershell
+Add-Content $PROFILE "`n`$env:Path += ';C:\Program Files\LLVM\bin'`n`$env:CC='clang-cl'`n`$env:CXX='clang-cl'"
+```
+
+(If the profile doesn't exist yet: `New-Item -ItemType File -Path $PROFILE -Force`. Then open a
```
vlad-ivanov-name: does not belong here probably

docs/src/contributing/windows.md
```diff
@@ -0,0 +1,134 @@
...
+
+```powershell
+cargo build --release -p josh-proxy -p josh-cli
+```
+
+Binaries land in `target\release\`: `josh-proxy.exe`, `josh.exe`, `josh-filter.exe`.
+
+Clone to a neutral path: some agent-tool workspaces inject ephemeral credentials into a
+repository's local config, and once those expire `git pull` fails with an auth error even on a
+public repository, because a stale credential beats no credential.
```
vlad-ivanov-name: this is specific to the local setup, doesn't need to be here

docs/src/contributing/windows.md
```diff
@@ -0,0 +1,134 @@
...
+target\release\josh-proxy.exe --local C:\josh-cache --remote=https://github.com --port=8000 --no-background
+```
+
+Then, from another terminal, clone anything through a filter:
+
+```powershell
+git clone "http://127.0.0.1:8000/josh-project/josh.git:prefix=josh.git" filtered-test
+```
+
+## Testing
```
vlad-ivanov-name: i think we can skip the entire windows testing section for now, just mention that josh compose run doesn't work on windows

docs/src/contributing/windows.md
```diff
@@ -0,0 +1,134 @@
...
+`go` and `curl` on PATH.
+
+## Troubleshooting
+
+Errors seen while getting josh working on Windows, with the short version of each diagnosis.
+
+| Symptom | Cause and fix |
+|---|---|
+| `LNK1181` from `lib.exe` after thousands of C warnings | ARM64 only: `cl.exe` cannot assemble aws-lc's GNU-syntax ARM assembly, so the objects it archives were never written. Install LLVM and set `CC`/`CXX` to `clang-cl` (above). |
+| Build output floods the scrollback | A failing build script dumps its whole log. Capture it: `cargo build ... 2>&1 \| Out-File build.log`, then `Get-Content build.log -Tail 80`. |
```
vlad-ivanov-name: most of these are local setup issues; some are fixed in this PR. whole section can be removed, or only the comment about LNK1181 left

docs/src/contributing/windows.md
```diff
@@ -0,0 +1,134 @@
...
+## Limitations
+
+- SSH serving is unix-only by design of the unix-socket handoff; HTTP serving is the supported
+  path on Windows. The `josh-ssh-shell` crate does not build here — it uses unix sockets, fifos
+  and raw file descriptors throughout — so build the binaries you need (`-p josh-proxy
+  -p josh-cli`) rather than the whole workspace.
+- `josh compose` (container runs) is untested on Windows; the uid/gid the containers run as
+  falls back to `1000:1000`, mirroring the default user of the VM podman runs containers in.
+- Windows path handling has test coverage (the `-PathForms` cases) but little real-world
+  mileage; treat long-path and mapped-drive setups as frontier territory.
```
vlad-ivanov-name: whole of this can be replaced with:

* SSH is not supported on Windows
* Windows support is experimental

josh-cli/src/bin/josh.rs
```diff
@@ -342,6 +344,15 @@ fn to_absolute_remote_url(url: &str) -> anyhow::Result<String> {
             .display()
             .to_string();
 
+        // On Windows canonicalize returns extended-length paths, which git rejects inside
+        // file:// URLs (issue #2288). \\?\UNC\server\share keeps its authority, so it becomes
+        // file://server/share/...; \\?\C:\... becomes file:///C:/... .
+        #[cfg(windows)]
+        let path = match path.strip_prefix(r"\\?\UNC\") {
```
vlad-ivanov-name: can you try if https://crates.io/crates/dunce works? it's already in the cargo lock

josh-compose-podman/src/lib.rs
```diff
@@ -38,12 +38,20 @@ impl Default for PodmanRuntime {
...
+#[cfg(unix)]
 fn host_uid_gid() -> (u32, u32) {
     let uid = unsafe { libc::getuid() };
     let gid = unsafe { libc::getgid() };
     (uid, gid)
 }
 
+/// On non-unix hosts there is no uid/gid to mirror; podman runs containers inside a VM whose
+/// default user is 1000:1000, so that is the closest equivalent.
+#[cfg(not(unix))]
```
vlad-ivanov-name: should be #[cfg(windows)], not unix is not necessarily windows

josh-core/src/filter/tree.rs
```diff
@@ -820,6 +820,14 @@ fn replace_child_inner(
     mode: i32,
     tree_oid: git2::Oid,
 ) -> anyhow::Result<git2::Oid> {
+    // A tree entry name is never empty. A path component that has no byte
+    // representation on this platform arrives here as one, so refuse rather
+    // than writing a tree git will reject.
```
vlad-ivanov-name: i am confused how this can happen 🤔

josh-gix-ext/src/lib.rs
```diff
@@ -23,6 +23,28 @@ use std::collections::HashMap;
...
+    std::os::unix::ffi::OsStrExt::as_bytes(c)
+}
+
+/// The raw bytes of a path component, for matching against git tree entry names.
+///
+/// Paths are Unicode here and git tree names are conventionally UTF-8, so the UTF-8 encoding
+/// is the equivalent of the unix byte representation. A component that is not valid Unicode
+/// yields an empty slice, which matches nothing on lookup — tree entry names are never empty —
+/// and is refused by the tree-writing path rather than written out.
+#[cfg(not(unix))]
```
vlad-ivanov-name: #[cfg(windows)]

josh-gix-ext/src/lib.rs
```diff
@@ -23,6 +23,28 @@ use std::collections::HashMap;
...
+pub fn component_bytes(c: &std::ffi::OsStr) -> &[u8] {
+    c.to_str().map(str::as_bytes).unwrap_or(&[])
```
vlad-ivanov-name: yeah i don't think this unwrap is good here; i'd prefer to panic. can it ever panic?

josh-proxy/src/bin/josh-proxy.rs
```diff
@@ -177,7 +177,19 @@ async fn run_proxy(args: josh_proxy::cli::Args) -> anyhow::Result<i32> {
...
+    let socket = socket2::Socket::new(
+        socket2::Domain::IPV6,
+        socket2::Type::STREAM,
+        Some(socket2::Protocol::TCP),
+    )?;
+    socket.set_only_v6(false)?;
+    socket.bind(&addr.into())?;
+    socket.listen(1024)?;
+    socket.set_nonblocking(true)?;
+    let listener = tokio::net::TcpListener::from_std(socket.into())?;
```
vlad-ivanov-name: let's factor out this whole thing into a make_listener function

JohnCampionJr:
Some context that didn't belong in the commit message, and a few offers.

**How this was tested.** `josh compose run` needs podman, which isn't available on Windows, so the port was validated with two hermetic functional suites driving the built binaries against a local git-HTTP server — no network, fresh temp directories ([rigsmith/josh-binaries](https://github.com/rigsmith/josh-binaries)):

- proxy: boot/teardown, filtered clone (`:prefix`), pinned-SHA fetch (`@sha`), reverse-filter push asserting the re-rooted commit's parent is the upstream tip, warm-cache reuse across a proxy restart
- cli: `josh-filter`, `josh clone` into a relative directory (the #2288 trigger shape), `josh changes pull`, `josh push --base` asserting exact parentage

Both pass on Windows x86-64 (GitHub runners) and Windows ARM64 (Windows 11 on Apple silicon), the latter also with relative, space-laden, junction and `subst`-mapped cache paths. Both pass on macOS arm64 and Linux, and `cargo check --workspace --all-targets` is clean on unix.

**Two things I'd split out if you prefer.** The dual-stack listener is a cross-platform behavior change (explicit `IPV6_V6ONLY=0` rather than the OS default), not strictly a Windows fix — happy to move it to its own PR. And `josh-ssh-shell` gating is left out deliberately, so `cargo build --workspace` still fails on Windows while `-p josh-proxy -p josh-cli` succeeds; that crate uses unix sockets, fifos, raw fds and `libc::isatty` directly and never built there, so fixing it is mechanical but noisy and orthogonal. Say the word and I'll do it separately.

**Two behaviors I noticed and did not change**, in case they're intentional: `get_head` falls back to `refs/heads/master` when a remote's HEAD is unborn, so a filtered clone comes back empty but "successful"; and `josh push` without `--base` writes a parentless commit where the proxy path refuses outright. Happy to file either as an issue.

**Offer.** The CI that builds and functionally tests these binaries per platform (including the Windows path-form regressions for #2288) lives in that same repo. If a Windows job here would be useful, I'd gladly contribute it — master grew a new unix-only assumption in `josh-search`/`josh-gix-ext` in the weeks after r26.07.19, so without a build on the platform this will drift back to broken.

josh-proxy/src/bin/josh-proxy.rs
```diff
@@ -285,14 +297,19 @@ fn main() -> std::process::ExitCode {
     // The update hook will then make a http request back to the main
     // process to do the actual computation while taking advantage of the
     // cached data already loaded into the main process's memory.
+    // Hook invocations are recognized by argv[0] (the unix hooks are symlinks to this binary)
+    // or by JOSH_HOOK, which the shim-script hooks set on platforms where symlinking the
+    // executable isn't an option.
+    let josh_hook = std::env::var("JOSH_HOOK").unwrap_or_default();
```
vlad-ivanov-name: i feel like this leaves some space for potential security problems so let's 1) gate it to windows only 2) wrap it so that env var is only read once. something along the lines of

```
fn get_hook() -> Option<String> {
  #[cfg(windows)]
  {
    static JOSH_HOOK: LazyLock<Option<String>> = LazyLock::new(|| {
        std::env::var("JOSH_PROXY_HOOK").unwrap_or_default()
    });
    &*JOSH_HOOK.cloned()
  }

  #[cfg(not(windows))
  {
    None
  }
}
```

josh-proxy/src/service.rs
```diff
@@ -399,6 +399,39 @@ fn create_repo_base(path: &PathBuf) -> anyhow::Result<crate::shell::Shell> {
...
+fn install_hook(josh_executable: &std::path::Path, hook: &std::path::Path) -> std::io::Result<()> {
+    std::os::unix::fs::symlink(josh_executable, hook)
+}
+
+/// Install the josh executable as a git hook, as a shell shim: symlinking an executable may
+/// require elevated privileges here, and git runs hooks through sh anyway. Hook dispatch in the
+/// binary keys off argv[0], which a shim cannot fake, so the shim passes the hook name in
+/// JOSH_HOOK. The path is absolutized (git runs hooks with GIT_DIR as the working directory)
+/// and single-quoted so no part of it is taken as a shell expansion.
+#[cfg(not(unix))]
```
vlad-ivanov-name: cfg(windows)

josh-proxy/src/service.rs
```diff
@@ -399,6 +399,39 @@ fn create_repo_base(path: &PathBuf) -> anyhow::Result<crate::shell::Shell> {
...
+fn install_hook(josh_executable: &std::path::Path, hook: &std::path::Path) -> std::io::Result<()> {
+    let name = hook
+        .file_name()
+        .map(|n| n.to_string_lossy())
+        .unwrap_or_default();
+    // git runs hooks with GIT_DIR as the working directory, so a relative path would resolve
+    // under the overlay repo; make it absolute. Single-quote the result (escaping any quote)
+    // so no part of the path is taken as a shell expansion.
+    let exe = std::path::absolute(josh_executable)
```
vlad-ivanov-name: `dunce` can be used here as well?

josh-proxy/src/service.rs
```diff
@@ -399,6 +399,39 @@ fn create_repo_base(path: &PathBuf) -> anyhow::Result<crate::shell::Shell> {
...
+        .unwrap_or_else(|_| josh_executable.to_path_buf())
+        .to_string_lossy()
+        .replace('\\', "/")
+        .replace('\'', r"'\''");
+    std::fs::write(
```
vlad-ivanov-name: minor nit but since it's agent looking at those comments anyway: this could be more readable as a sequence of write!() / writeln!() calls or string appends, one shell line per one rust code line

vlad-ivanov-name (commented):
thank you for the contribution!

some notes:

1) please run the tidy-comments skill from the repo. looks like code is authored with claude -- with claude you have to command it to be really strict about comments

2) ideally we would want windows CI on github actions. as a first step to include in this PR -- maybe just building josh, without running .t tests? we could still run unit tests of the crates with e.g. nextest. if some crates don't pass tests, it's fine if it's just one or two, still better than nothing

3) it's not entirely clear what the state of podman is in this PR, some places mention podman VM on windows, some mention it didn't work; if it didn't work, it's better to just exit at runtime when trying to run compose

docs/src/contributing/windows.md
JohnCampionJr: obviously, sorry.

JohnCampionJr:
Thanks for the thorough review — all of it addressed, and running the tests on Windows turned up a real bug that I would not have found otherwise.

**1) tidy-comments.** Ran it. Dropped the comments that only restated code, shortened the rest to the "why", and removed the two doc comments I had added purely to satisfy a docstring-coverage bot.

**2) Windows CI.** Added as `.github/workflows/rust-windows.yml`, so `rust.yml` is untouched. It builds `-p josh-proxy -p josh-cli` (the workspace does not build there — see the scope note below), runs the unit tests of the six crates that build there with nextest, and then runs functional tests against the built binaries, since `.t` is out of reach without podman:

- `tests/windows/cli.sh` — `josh-filter`, `josh clone` into a relative directory, `josh changes pull`, `josh push --base`, against a local repository. No server, no network.
- `tests/windows/proxy.sh` — filtered clone, pinned-SHA fetch, reverse-filter push (asserting the pushed commit's parent is the upstream tip), and reuse of the `--local` cache across a proxy restart. Repeated with relative and space-laden cache directories.

`josh-proxy` only accepts an http(s)/ssh upstream, so the proxy tests need a git server. Rather than add a dependency, `tests/windows/serve-git.ps1` hosts git's own `http-backend` as CGI behind `System.Net.HttpListener`. The job installs nothing that Windows does not already ship. Every step is a hard gate — no `continue-on-error` — because all of it passes.

**3) podman.** You were right that the earlier version was muddled. `josh compose` is now refused up front on Windows with a plain error, and the uid/gid helper there is `unreachable!()` instead of an invented `1000:1000`. The docs say compose does not run there, which is also why the `.t` suites cannot.

**Inline comments:** `#[cfg(windows)]` everywhere instead of `not(unix)`; `dunce` works and simplified the URL handling (it also removed the need for the separate `\\?\UNC\` case, since dunce returns ordinary paths); `make_listener` factored out; `JOSH_PROXY_HOOK` gated to Windows and read once through `LazyLock`, per your sketch; the shim is built with `writeln!` calls, one shell line each; the docs are down from 148 lines to 42, with the binaries link gone.

On `tree.rs` — your confusion was justified, and it is gone. `component_bytes` used to map an unrepresentable component to an empty slice, and that guard was catching the consequence. It now panics instead, which cannot reach the tree writer, so the guard is removed.

**What testing on Windows found, beyond compiling:**

- **`.gitmodules` was written with CRLF on Windows.** `gix-config` writes the platform newline, and that content becomes a blob — so the same filter applied to the same input produced a different blob, tree and commit hash on Windows than on Linux. Fixed by normalizing to LF. This one affects the correctness of filtering itself, not just Windows usability.
- A test identifying a ref file by inode, and a test building a ref path containing `aux` — a reserved Windows device name, which the OS cannot represent at all. Both gated to unix; the device-name limitation is documented, since it means an upstream whose path contains one cannot be mirrored on Windows.

**Scope note:** `josh-ssh-shell` is still untouched and still does not build on Windows — it uses unix sockets, fifos, raw fds and `libc::isatty` directly, independent of this PR — so `cargo build --workspace` does not succeed there while `-p josh-proxy -p josh-cli` does. Happy to gate that crate separately if you want the workspace green.

docs/src/contributing/windows.md
JohnCampionJr: Removed — the whole intro is now two sentences.

docs/src/contributing/windows.md
JohnCampionJr: Trimmed to the winget list plus one line saying Git for Windows is needed at runtime, not just to clone (josh shells out to git, and the hooks it installs are sh shims).

docs/src/contributing/windows.md
JohnCampionJr: Removed.

docs/src/contributing/windows.md
JohnCampionJr: Removed.

docs/src/contributing/windows.md
JohnCampionJr: The section is gone. The docs now just say `josh compose run` is not supported on Windows, so the repository's own suite does not run there. The functional tests that took its place live in `tests/windows/` and run in CI.

docs/src/contributing/windows.md
JohnCampionJr: Removed, keeping only the LNK1181 note, since that one is not a local misconfiguration: on ARM64 the aws-lc-sys assembly needs clang-cl, and the failure surfaces late and unhelpfully in `lib.exe`.

docs/src/contributing/windows.md
JohnCampionJr: Replaced with exactly that, plus one more: an upstream whose path contains a reserved Windows device name (`aux`, `con`, `nul`, `com1`..`lpt9`) cannot be mirrored, because the namespace becomes a path. That surfaced as a failing test, now gated to unix.

docs/src/contributing/windows.md
JohnCampionJr: Removed. The functional tests are contributed here instead, in `tests/windows/`, so nothing in the docs points at binaries you have not built.

josh-cli/src/bin/josh.rs
JohnCampionJr: `dunce` works, and it simplified the change: it returns an ordinary path where it can, so the extended-length prefix handling is gone entirely. What remains is turning backslashes into forward slashes and keeping the authority of a UNC path (`file://server/share/...`) while a drive path becomes `file:///C:/...`.

josh-compose-podman/src/lib.rs
JohnCampionJr: Changed to `#[cfg(windows)]` — you are right that not-unix is not necessarily Windows. Also reworked per your podman note: `josh compose` is now refused up front on Windows, so this helper is `unreachable!()` rather than inventing a uid/gid.

josh-core/src/filter/tree.rs
JohnCampionJr: Justified confusion — it cannot happen any more, and the guard is gone. It existed because `component_bytes` used to map an unrepresentable path component to an empty slice; that would have reached the tree writer as an empty entry name. It now panics instead, so the guard had nothing left to catch.

josh-gix-ext/src/lib.rs
JohnCampionJr: Changed to `#[cfg(windows)]`.

josh-gix-ext/src/lib.rs
JohnCampionJr: Agreed, and it now panics: `c.to_str().expect("path component is not valid Unicode")`. To your question — it can only fire on a path containing an unpaired surrogate, which cannot name anything in a git tree, so panicking is the honest response. Silently substituting something else would have corrupted both lookups and writes.

josh-proxy/src/bin/josh-proxy.rs
JohnCampionJr: Factored out as `make_listener`.

josh-proxy/src/bin/josh-proxy.rs
JohnCampionJr: Done, as you sketched: `hook_from_env()` with a `LazyLock<Option<String>>` inside `#[cfg(windows)]`, returning `None` on unix so the environment is never consulted there. Renamed the variable to `JOSH_PROXY_HOOK` to match your naming.

josh-proxy/src/service.rs
JohnCampionJr: Changed to `#[cfg(windows)]`.

josh-proxy/src/service.rs
JohnCampionJr: Yes — using `dunce::canonicalize` here too. It also covers the other half of the problem: git runs hooks with `GIT_DIR` as the working directory, so the path has to be absolute regardless.

josh-proxy/src/service.rs
JohnCampionJr: Rewritten as a sequence of `writeln!` calls, one shell line per Rust line. Also ran the tidy-comments skill over the whole change while I was in there.

JohnCampionJr:
The Windows job is in the PR now and green on both x86-64 and arm64: https://github.com/rigsmith/josh/actions/runs/32911292650

Two notes on it. It uses `cargo test` rather than nextest — the install action 403s under the restricted token a fork PR gets, and dropping it removed the failure mode entirely; happy to switch back if it runs fine with your token. And on arm64 it builds the C dependencies with clang-cl, which the runner image already provides, since aws-lc-sys' ARM assembly is GNU-syntax.

vlad-ivanov-name:
Thanks for the fixes! I will pick up the PR now; you don't need to address follow up comments.

.github/workflows/rust-windows.yml
```diff
@@ -0,0 +1,77 @@
+name: rust-windows
```
vlad-ivanov-name, line 1: we need to skip windows CI on draft PRs.

.github/workflows/rust-windows.yml
```diff
@@ -0,0 +1,77 @@
...
+    runs-on: ${{ matrix.os }}
+    name: Windows ${{ matrix.arch }}
+    if: github.event_name != 'pull_request' || github.event.pull_request.draft == false
+    steps:
+    - name: Checkout
+      uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+    - name: Install Rust toolchain
+      uses: dtolnay/rust-toolchain@6c977a6ca4077a0ceb28ffbe03f59d46e9ac8772 # stable
+      with:
+        toolchain: stable
```
vlad-ivanov-name, line 35: we should pin rust version matching the rust toolchain toml, it's fine if it's copied here instead of derived from toml

.github/workflows/rust-windows.yml
```diff
@@ -0,0 +1,77 @@
...
+    branches: [ '**' ]
+    types: [opened, synchronize, reopened, ready_for_review]
+  merge_group:
+
+env:
+  CARGO_TERM_COLOR: always
+  # aws-lc-sys (rustls' default provider) assembles with NASM on x86-64; the
+  # crate ships prebuilt objects behind this switch, so the runner needs no
+  # extra install. ARM64 hosts need clang-cl instead.
+  AWS_LC_SYS_PREBUILT_NASM: 1
```
vlad-ivanov-name, line 16: can we simplify this to always use clang-cl to avoid this weird branching?

.github/workflows/rust-windows.yml
```diff
@@ -0,0 +1,77 @@
...
+    - name: Install Rust toolchain
+      uses: dtolnay/rust-toolchain@6c977a6ca4077a0ceb28ffbe03f59d46e9ac8772 # stable
+      with:
+        toolchain: stable
+    # aws-lc-sys' ARM assembly is GNU-syntax, which MSVC cannot assemble: the
+    # build then fails late in lib.exe, archiving objects that were never
+    # written. clang-cl assembles it and stays ABI-compatible with the MSVC
+    # toolchain the Rust side links with. The runner image ships it; the step
+    # reports what it found, and warns loudly if it ever has to install one.
+    - name: Use clang-cl for C dependencies
```
vlad-ivanov-name, line 41: we need to simplify this.

1) we need to install llvm of exact pinned version, always
2) we don't need to check what host has to offer -- we don't want to depend on it.

josh-cli/src/bin/josh.rs
```diff
@@ -336,12 +336,20 @@ fn to_absolute_remote_url(url: &str) -> anyhow::Result<String> {
...
             .with_context(|| format!("Failed to resolve path {}", url))?
             .display()
             .to_string();
 
+        // A UNC path keeps its authority (file://server/share/...); a drive path does not.
+        #[cfg(windows)]
+        let path = match path.strip_prefix(r"\\") {
+            Some(unc) => unc.replace('\\', "/"),
+            None => format!("/{}", path.replace('\\', "/")),
+        };
```
vlad-ivanov-name, line 351: isn't this covered by dunce?

josh-cli/src/bin/josh.rs
```diff
@@ -336,12 +336,20 @@ fn to_absolute_remote_url(url: &str) -> anyhow::Result<String> {
     {
         Ok(url.to_owned())
     } else {
-        // For local paths, make them absolute
-        let path = std::fs::canonicalize(url)
+        // dunce, not std: on Windows std::fs::canonicalize returns an extended-length path
+        // (\\?\C:\...), which git rejects inside a file:// URL (issue #2288).
+        let path = dunce::canonicalize(url)
```
vlad-ivanov-name, line 341: i would rather prefer to call dunce on windows only, via cfg

josh-core/src/submodules.rs
```diff
@@ -104,7 +104,10 @@ pub fn update_gitmodules(
         .write_to(&mut output)
         .context("Failed to write gitmodules")?;
 
-    String::from_utf8(output).context("Invalid UTF-8 in gitmodules")
+    let content = String::from_utf8(output).context("Invalid UTF-8 in gitmodules")?;
+    // gix-config writes the platform's newline, but this ends up in a blob: the same
```
vlad-ivanov-name, line 108: we should check gix-config configuration knobs -- see if we can avoid awkward string manipulation below

josh-proxy/src/service.rs
```diff
@@ -399,6 +399,38 @@ fn create_repo_base(path: &PathBuf) -> anyhow::Result<crate::shell::Shell> {
...
+        .file_name()
+        .map(|n| n.to_string_lossy())
+        .unwrap_or_default();
+    // Absolute, because git runs hooks with GIT_DIR as the working directory; quoted, so no
+    // part of the path is a shell expansion.
+    let exe = dunce::canonicalize(josh_executable)
+        .unwrap_or_else(|_| josh_executable.to_path_buf())
+        .to_string_lossy()
+        .replace('\\', "/")
+        .replace('\'', r"'\''");
```
vlad-ivanov-name, line 424: is this even necessary after dunce?

josh-proxy/src/service.rs
```diff
@@ -399,6 +399,38 @@ fn create_repo_base(path: &PathBuf) -> anyhow::Result<crate::shell::Shell> {
...
+
+    let mut script = String::new();
+    let _ = writeln!(script, "#!/bin/sh");
+    let _ = writeln!(script, "JOSH_PROXY_HOOK={name}");
+    let _ = writeln!(script, "export JOSH_PROXY_HOOK");
+    let _ = writeln!(script, "exec '{exe}' \"$@\"");
```
vlad-ivanov-name, line 430: nit: should be raw string

vlad-ivanov-name:
Unfortunately because this is a cross-org PR (as opposed to coming from a user fork), I can't make changes here, this is a known github bug.

I opened a new PR and will continue work there:

https://github.com/josh-project/josh/pull/2539

JohnCampionJr:
> Unfortunately because this is a cross-org PR (as opposed to coming from a user fork), I can't make changes here, this is a known github bug.
> 
> I opened a new PR and will continue work there:
> 
> #2539

I always forget that, my bad.  Thanks for taking it across the finish line :)

vlad-ivanov-name:
Superseded by https://github.com/josh-project/josh/pull/2539

@JohnCampionJr thanks again for the contribution!

One note on the scope of Windows support: while we did merge proxy tests as well, most of the time we encourage people to use `josh` CLI when building workflows locally -- it simplifies some operations as you no longer need a service running in background. Proxy is intended for deployments where it's shared with multiple users, and in cases where you want to point an existing Git client at an URL of the proxy.

JohnCampionJr:
I'm actually using the proxy for a workflow so I appreciate you leaving it in :) Thanks for taking the PR!

vlad-ivanov-name:
Yes, quite a few people do. So this is more about the direction where the project is going. rust-lang's josh-sync will eventually switch to CLI as well.

JohnCampionJr:
That was my model.  Once that can switch to CLI, I can too :)

PR #2539, opened 2026-08-26
Support building and running josh on Windows

vlad-ivanov-name:
Make josh-proxy and the josh / josh-filter CLIs build and run on Windows (issue #2235). Unix behavior is unchanged: unix-only code moves behind cfg gates, and the few places where Windows differs get their own implementation:

- Path handling: tree-entry name matching no longer assumes OsStr is bytes, and local paths become valid file:// URLs (issue #2288).
- josh-proxy: hooks install as a small sh shim instead of a symlink (Windows symlinks need elevated privileges), the listener binds dual-stack, and SSH serving reports "unsupported on this platform" instead of failing to build.
- josh-core: the unused update_gitmodules helper is removed. Nothing called it, and its gix-config based writer was the one piece of submodule code with platform-dependent output.
- josh-cli: `josh compose` refuses to run on Windows, where podman is unsupported.

CI: a new rust-windows workflow (rust.yml untouched) builds on x86-64 and arm64, runs the unit tests of the crates that build there, and runs functional tests from tests/windows/ (all PowerShell): the CLI against a local repository, and josh-proxy against a local git server through a filtered clone, a pinned-SHA fetch, a reverse push, and cache reuse across a restart. The job uses sccache with the shared R2 backend and a pinned LLVM toolchain, skips draft PRs, and runs without secrets on forks.

Docs: docs/src/contributing/windows.md covers setup, build, and limitations.

Tested on Windows x86-64 and ARM64, macOS, and Linux.


josh-core/src/submodules.rs
```diff
@@ -104,7 +104,13 @@ pub fn update_gitmodules(
...
         .context("Failed to write gitmodules")?;
 
-    String::from_utf8(output).context("Invalid UTF-8 in gitmodules")
+    let content = String::from_utf8(output).context("Invalid UTF-8 in gitmodules")?;
+    // gix-config's File::write_to reuses the input's newlines and falls back to the
+    // platform's when the input has none; its only newline knob (parse::format::
+    // Options) belongs to the separate re-formatting API, not the mutation API used
+    // here. This ends up in a blob, so the same filter has to produce the same object
+    // on every platform.
+    Ok(content.replace("\r\n", "\n"))
```
christian-schilling: I'm not sure this is right. We had serveral instances of normalizing (gpgsig comes to mind) where normalizing turned out to be wrong.

josh-gix-ext/src/lib.rs
```diff
@@ -4,6 +4,21 @@ use std::collections::HashMap;
...
+pub fn component_bytes(c: &std::ffi::OsStr) -> &[u8] {
+    std::os::unix::ffi::OsStrExt::as_bytes(c)
+}
+
+/// The raw bytes of a path component, for matching against git tree entry names: their UTF-8
+/// encoding. Panics on a component that is not valid Unicode, which cannot name a tree entry.
+#[cfg(windows)]
+pub fn component_bytes(c: &std::ffi::OsStr) -> &[u8] {
+    c.to_str()
+        .expect("path component is not valid Unicode")
```
christian-schilling, line 18: This also does not seem right. How to those two differ?

vlad-ivanov-name: windows uses UTF-16 codepairs to represent paths. there is no direct conversion to bytes unless we choose an encoding. here we choose UTF-8. this would panic on unpaired UTF-16 surrogate. any "normal" software is not supposed to produce it. but it's possible to do this on purpose, or by using some broken software. here we are assuming git won't do that.

josh-core/src/submodules.rs
vlad-ivanov-name: i think this means some of the tests will have platform specific expectations. what about detecting endline used in gitmodules_content and falling back to platform default?

josh-gix-ext/src/lib.rs
vlad-ivanov-name: or in other words we either have to panic here or be lossy.

josh-core/src/submodules.rs
vlad-ivanov-name: removed the fn entirely 🪓

josh-gix-ext/src/lib.rs
christian-schilling: ok, panic is better than lossy :)

christian-schilling (approved):

PR #2558, opened 2026-09-01
Rebuild missing cached trees

kyle-oai:
Intermediate tree caches can reference Git objects that are absent from the current repository. Garbage collection or reuse across repositories can leave these entries behind. Returning a missing tree ID then causes filtering to fail.

Treat missing cached path, inverse, glob, pattern, and trigram trees as cache misses. Rebuild the trees and replace stale entries with the rebuilt results.

Validation:

- `cargo fmt --all -- --check`
- `cargo test --locked -p josh-core -p josh-search --lib`: 93 tests passed (88 core, 5 search).
- Regression tests cover each affected tree cache.

This is the tree-recovery part of #2557, based directly on `master`.

### DIFF
--- a/josh-core/src/cache/transaction.rs
+++ b/josh-core/src/cache/transaction.rs
@@ -1177,19 +1177,21 @@ impl Transaction {
     }
 
     pub fn insert_paths(&self, tree: (gix_hash::ObjectId, String), result: gix_hash::ObjectId) {
-        PATHS_MAP.write().unwrap().entry(tree).or_insert(result);
+        PATHS_MAP.write().unwrap().insert(tree, result);
     }
 
     pub fn get_paths(&self, tree: (gix_hash::ObjectId, String)) -> Option<gix_hash::ObjectId> {
-        PATHS_MAP.read().unwrap().get(&tree).cloned()
+        let cached = PATHS_MAP.read().unwrap().get(&tree).copied()?;
+        self.odb().contains(cached).then_some(cached)
     }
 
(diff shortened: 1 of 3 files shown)


christian-schilling:
Looks reasonable. Since this is very much on a hot path I'll do some benchmarks tomorrow...

christian-schilling (changes requested):
Running our benches showed that this causes a 5-10% perf regression on the relevant operations (globs, indexing,...). 
So while thinking about the tradeoffs along the lines of "this only affects some filter and the hit is not that big...we might want to make that trade to avoid failing filtering..." I realized: This is most likely not the right way to fix either problem I can think of motivating this:
1) Working with multiple repos/ODBs in the same process.
First: I'd avoid doing it but let's say you really want to. The better fix would be get rid of the process scoped in memory caches entirely and instead move all of them to the transaction scope (we already have some of those). This is long overdue, would reduce long term memory consumption and solve the "cross repo missing objects" issue.
I have benchmarked this. It does regress very short walks (which are fast anyway) but has almost no impact on long ones, so this is a good trade.

2) Concurrent GC
We don't support concurrent GC on the repo while a josh process is running. Repack yes. GC/prune no. (see repack options used in josh-proxy)
Even if we wanted to support this, this PR does not fix the problem: There is still a race window between the cache lookup + existence check and the actual object read later.
So *if* concurrent GC is really a requirement we'd need to design a solution with actual guarantees instead of just making the error a bit less likely by shortening the time window for the race.

christian-schilling:
@kyle-oai We merged https://github.com/josh-project/josh/pull/2576 which removes long lasting in process caches in favor of transaction scoped ones and could already solve your staleness problem.

PR #2559, opened 2026-09-01
Add optional caching of requested commit endpoints

kyle-oai:
Sparse distributed-cache sampling can skip the last commit requested by a caller. A fresh process must then walk past that endpoint when filtering later commits.

Add `apply_to_commit_with_options` with `ApplyToCommitOptions { force_cache: true }`. Writable backends store the requested input-to-output commit mapping even when sampling would skip it. Ordinary history traversal reads those entries, so filtering a later commit can reuse its cached parent. Existing `apply_to_commit` calls keep default sampling. Read-only backends still ignore writes.

Validation:

- `cargo fmt --all -- --check`
- `cargo test --locked -p josh-core --lib`: 90 tests passed.
- Regression tests cover endpoint persistence, cold-cache parent reuse, local-cache promotion, null outputs, read-only backends, and internal filters.

This is the endpoint-caching part of #2557, based directly on `master`.

### DIFF
--- a/josh-core/src/cache/distributed.rs
+++ b/josh-core/src/cache/distributed.rs
@@ -207,8 +207,9 @@ impl DistributedCacheBackend {
 // is used in addition to the sparse cache.
 // The sparse cache is mostly only used for initial "cold starts" or longer "catch up".
 // For incremental filtering it's fine re-filter commits and rely on the local "dense" cache.
-// Only the sample points are persisted; see `HistoryGraphHint::is_sample_point` for the
+// Ordinary writes persist only sample points; see `HistoryGraphHint::is_sample_point` for the
 // invariant that bounds how far a walk runs before it hits one.
+// Explicitly requested endpoints can also be persisted and reused during history traversal.
 //
 // To additionally limit the size of the trees the cache is also sharded by sequence
 // number in groups of 10000. Note that this does not limit the number of entries per bucket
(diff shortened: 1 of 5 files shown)


josh-core/src/filter/mod.rs
```diff
@@ -374,12 +374,34 @@ pub fn apply_to_commit(
...
+    options: ApplyToCommitOptions,
 ) -> anyhow::Result<gix_hash::ObjectId> {
     let filter = opt::optimize(filter);
+    let force_cache =
+        options.force_cache && filter != sequence_number() && filter != reachable_roots();
     loop {
         let filtered = apply_to_commit2(filter, commit, transaction)?;
 
         if let Some(id) = filtered {
+            if force_cache {
```
christian-schilling, line 402: It looks to me like this can be done simpler: Since the "forced" store happens for the final result the caller could just
`let r = apply_to_commit(..); transaction.insert_forced(...,r)` and no new function variant would be needed

josh-core/src/filter/mod.rs
```diff
@@ -2684,6 +2706,174 @@ mod tests {
         .unwrap()
     }
 
+    fn unsampled_endpoint(
```
christian-schilling, line 2709: Not sure what the term "endpoint" means 🤔

josh-core/src/cache/distributed.rs
```diff
@@ -236,17 +237,12 @@ impl CacheBackend for DistributedCacheBackend {
...
         if filter == filter::sequence_number() || filter == filter::reachable_roots() {
             return Ok(None);
         }
-        // Tree-keyed records were written by whichever indexing commit was eligible,
-        // so gating reads on the reader's eligibility would hide them from the
-        // (usually non-sampled) commits that reuse them.
-        if !tree_keyed && !hint.is_sample_point() {
-            return Ok(None);
-        }
+        // Reads also honor forced endpoints and tree-keyed records from sampled commits.
```
christian-schilling, line 245: This could be a *severe* performance regression: The current code avoids most ODB reads, while the change will cause a read on every call.
Will benchmark tomorrow. (You can of course also do that...)

christian-schilling:
We very recently changed the sampling strategy for the distributed cache: https://github.com/josh-project/josh/pull/2492
This was exactly to fix the `find_known` search to walk past sample points on branching histories.
Did you check that whatever problem you encountered with the sparse cache sampling is not already addressed by that change (latest master) or could be easily solved by adjusting the sampling ratio?

christian-schilling (changes requested):

PR #2596, opened 2026-09-13
Publish changes to links

vlad-ivanov-name:
Publish changes to links

"josh changes publish" now additionally publishes to every configured
link: the published changes are projected through each link's filter and
pushed to the link's remote as the usual @changes/@base/@heads refs, and
PRs are created when the link's forge is GitHub.

Changes that become empty under a link's filter (e.g. a commit touching
only paths outside the link's view) are skipped for that link. Link push
failures abort the publish.

A link's forge is stored in link.josh as a "forge" meta key, set via
"josh link add --forge/--no-forge" and defaulting to guessing from the
URL, mirroring remote configuration. The Forge enum moves from
josh-changes to josh-view so links can carry it; josh-changes re-exports
it for existing users.

A link's tracked ref is required to be fully qualified
(e.g. refs/heads/main), enforced both when writing and when reading
link.josh, so the ref namespace is unambiguous. Relaxing this later is
easier than tightening it.

Publishing to links is gated behind JOSH_EXPERIMENTAL_FEATURES=1, like
the "josh link" CRUD commands. Gerrit links can be stored but error out
at publish time for now.

Change: publish-to-links
Assisted-By: kimi-code/k3


tests/cli/link.t
```diff
@@ -70,7 +70,43 @@ Remove links.
...
+
+A forge can be recorded explicitly; for a non-GitHub URL the guess is empty.
+
+  $ josh link add ${TESTTMP}/remote :/sub1 --id gh --forge github | sed "s|${TESTTMP}|\${TESTTMP}|g"
+  Added link 'gh': ${TESTTMP}/remote master :/sub1
+
+  $ git show refs/josh/links/gh:link.josh | sed "s|${TESTTMP}|\${TESTTMP}|g"
+  :~(
+      forge="github"
+      tracked-ref="master"
```
christian-schilling: We might want tracked refs to be required to be fully qualified always. Thoughts about that?
One thought is that we can always make this more "relaxed" later.

vlad-ivanov-name: Done

christian-schilling (approved):

PR #2597, opened 2026-09-13
Sync and evaluate admission conditions for changes

vlad-ivanov-name:
Sync and evaluate admission conditions for changes

Replace the coarse review_decision/check_status rollups stored in
PrData with precise per-reviewer and per-check maps, and fetch the
remote-scoped admission inputs (maintainers, required status checks
from repository rulesets) once per sync into a new gh_admission/
namespace on the changes ref, cached with a one-week TTL.

Admission evaluation is now a pure function of PrData and the cached
AdmissionData, replacing the never-wired AdmissionState. josh changes
list shows a mergeable column and josh changes show prints the
admission breakdown; josh-gui keeps its display via rollup helpers
derived from the new maps.

Change: sync-admission-conditions
Assisted-By: kimi-code/k3


tests/cli/changes_cli.t
```diff
@@ -135,9 +135,9 @@ directly:
   $ josh changes list
   Changes on Local [master]:
   
-  5ecf7d2  c2  D=  1  C=  0  V=         B change
-  28e4ce5  c1  D=  0  C=  2  V=approve  A change
-  b28ecba  c3  D=  0  C=  0  V=         C change
+  5ecf7d2  c2  D=  1  C=  0  V=         M=-    B change
```
christian-schilling, line 138: So the "admissable" case never appears in tests?

vlad-ivanov-name: I'll have to follow up with that. We don't have admission data for repos without forge right now.

christian-schilling (approved):

